PR #18, opened 2017-09-13
KokkosTools - no mpi version miss "}"

kyungjoo-kim:
Approved by Dan Ibanez


ibaned:
@kyungjoo-kim and I fixed the compilation of this tool for the case when the hackish `USE_MPI` macro is not defined, currently thats the way we get the tool to not use MPI. Just gonna go ahead and merge.

Issue #22, opened 2017-10-12
Weird (buggy?) behavior from the space-time-stack tool


stanmoore1:
I'm seeing some perplexing behavior from the `space-time-stack` tool. The d_sorted (particle:sorted) View in my code has no other references (reference count = 1 always) and is periodically reallocated:
<pre>
    if (d_particles.dimension_0() > d_sorted.dimension_0()) {
      d_sorted = t_particle_1d("particle:sorted",d_particles.dimension_0());
    }
</pre>
I can't understand how, but the d_sorted (particle:sorted) view appears twice in the `space-time-stack` tool output list for CUDA:
<pre>
MPI RANK WITH MAX MEMORY: 0
MAX BYTES ALLOCATED: 8574610364
HOST ALLOCATIONS:
================
  19.4% particle:particles_mirror
  2.3% grid:cells_mirror
  1.2% grid:cinfo_mirror
  0.7% particle:mlist_mirror
  0.2% emit/face:tasks_mirror
  0.2% collide:vremax_mirror
  0.2% collide:remain_mirror
  0.2% thermal/grid:vector_grid_mirror
  0.2% thermal/grid:tally_mirror
CUDA ALLOCATIONS:
================
  19.4% particle:sorted
  19.4% particle:particles
  19.4% particle:sorted
  11.5% particle:plist
  2.3% grid:cells
  1.2% grid:cinfo
  0.7% particle:mlist
  0.2% emit/face:tasks
  0.2% collide:vremax
  0.2% collide:remain
  0.2% thermal/grid:vector_grid
  0.2% thermal/grid:tally
</pre>

If I first destroy the d_sorted view like this:
<pre>
    if (d_particles.dimension_0() > d_sorted.dimension_0()) {
      d_sorted = t_particle_1d();
      d_sorted = t_particle_1d("particle:sorted",d_particles.dimension_0());
    }
</pre>
then d_sorted isn't listed twice in the CUDA output, but the overall high water mark doesn't change, and another copy of the particle:particles_mirror host view appears!
<pre>
MPI RANK WITH MAX MEMORY: 0
MAX BYTES ALLOCATED: 8638309940
HOST ALLOCATIONS:
================
  19.3% particle:particles_mirror
  19.3% particle:particles_mirror
  2.3% grid:cells_mirror
  1.2% grid:cinfo_mirror
  0.7% particle:mlist_mirror
  0.2% emit/face:tasks_mirror
  0.2% collide:vremax_mirror
[...]

Issue #20, opened 2017-10-11
Memory in SpaceTimeStack tool seems wrong


stanmoore1:
The memory in the `SpaceTimeStack` tool seems wrong. The high water mark from this tool (MAX BYTES ALLOCATED) doesn't match the other Kokkos memory high water mark tools, and the percentages listed don't add up to 100%.

@ibaned

stanmoore1:
Also, I think that there needs to be a separate high water mark for CUDA and Host memory spaces.

ibaned:
The not adding up to 100% is probably due to a lower cutoff threshold that omits entries with less than a few percent contribution. I'll need a reproducer of some kind to debug the difference in totals between tools.

ibaned:
also, the `hwm` tools use this call:
```
getrusage(RUSAGE_SELF, &sys_resources);
sys_resources.ru_maxrss
```
which is an estimate of max CPU memory used by any memory allocation calls.
`space-time-stack` only measures `Kokkos::View` allocations, nothing more.

stanmoore1:
Here is the output from space-time-stack. The host memory only adds up to 19.40% and the CUDA memory adds up to 79.70%. Are those two supposed to make 100% together? I don't think that makes sense since host and CUDA use totally separate memory.

<pre>
MPI RANK WITH MAX MEMORY: 0
MAX BYTES ALLOCATED: 10696989988
HOST ALLOCATIONS:
================
  15.5% particle:particles_mirror
  1.8% grid:cells_mirror
  1.0% grid:cinfo_mirror
  0.6% particle:mlist_mirror
  0.1% emit/face:tasks_mirror
  0.1% collide:vremax_mirror
  0.1% collide:remain_mirror
  0.1% thermal/grid:vector_grid_mirror
  0.1% thermal/grid:tally_mirror
CUDA ALLOCATIONS:
================
  15.5% particle:particles
  15.5% particle:sorted
  15.4% particle:particles
  10.4% particle:plist
  9.8% particle:plist
  9.2% particle:plist
  1.8% grid:cells
  1.0% grid:cinfo
  0.6% particle:mlist
  0.1% emit/face:tasks
  0.1% collide:vremax
  0.1% collide:remain
  0.1% thermal/grid:vector_grid
  0.1% thermal/grid:tally

END KOKKOS PROFILING REPORT.
</pre>

ibaned:
[...]

PR #23, opened 2017-10-12
Separate HWMs for each space

ibaned:
addresses feedback in issues
[#22, #20]


stanmoore1:
@ibaned works great, thanks

PR #24, opened 2017-11-01
Fix MPI issue in space-time-stack

stanmoore1:
Without this change, `space-time-stack` prints out lines for every single MPI rank.


stanmoore1:
@ibaned

ibaned:
Thanks @stanmoore1 !

PR #26, opened 2017-11-01
Add min and average to memory-hwm-mpi

stanmoore1:
When there are a lot of MPI ranks, it would be nice to know min, max, and average high water marks instead of just max.


stanmoore1:
Here is an example output:
```
Host process high water mark memory consumption: 25940 kB
  Max: 25940, Min: 24700, Ave: 25045 kB
```

stanmoore1:
@crtrott @nmhamster Can you take a look?

stanmoore1:
Bump?

PR #27, opened 2017-11-01
Add CPU host process memory hwm to kp_space_time_stack

stanmoore1:
@ibaned it would be really nice to have both Kokkos View and CPU host process memory high water marks in a single MPI compatible tool; sometimes non-Kokkos allocations can also be significant. I think the easiest way to do this is to add this to `space-time-stack` tool, so I've added this PR for your consideration. When I say `host process` hwm I mean that measured by:

```
getrusage(RUSAGE_SELF, &sys_resources);
sys_resources.ru_maxrss
```

### DIFF
--- a/src/tools/space-time-stack/kp_space_time_stack.cpp
+++ b/src/tools/space-time-stack/kp_space_time_stack.cpp
@@ -8,6 +8,7 @@
 #include <set>
 #include <cassert>
 #include <queue>
+#include <sys/resource.h>
 
 #define USE_MPI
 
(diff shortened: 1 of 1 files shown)


stanmoore1:
Here is an example output. It may be good to unify the output formatting and switch everything to a consistent unit, say MB.
```
Host process high water mark memory consumption: 25940 kB
  Max: 25940, Min: 24700, Ave: 25045 kB
```

ibaned:
Yep, this seems fine to me. We may change the prints later to make it more clear to users what the difference is between this total and the other one, but I'll go ahead and accept this now.
Thanks!

PR #28, opened 2017-11-09
Fix compile error with kp_space_time_stack.cpp

stanmoore1:
Clang on BG/Q errors out and won't compile `kp_space_time_stack.cpp` without this change, see Issue #29

### DIFF
--- a/src/tools/space-time-stack/kp_space_time_stack.cpp
+++ b/src/tools/space-time-stack/kp_space_time_stack.cpp
@@ -394,7 +394,7 @@ struct Allocations {
       MPI_Irecv(const_cast<char*>(s.data()), string_size, MPI_CHAR, min_max_rank, 42, MPI_COMM_WORLD, &request);
     }
     if (rank == min_max_rank) {
-      MPI_Send(s.data(), string_size, MPI_CHAR, 0, 42, MPI_COMM_WORLD);
+      MPI_Send(const_cast<char*>(s.data()), string_size, MPI_CHAR, 0, 42, MPI_COMM_WORLD);
     }
     if (rank == 0) {
       MPI_Wait(&request, MPI_STATUS_IGNORE);


stanmoore1:
@ibaned

ibaned:
Thanks for the mention in the comment, thats the first notification I've seen about this.

Actually, I think I've run into this before (the BG/Q has a bad MPI implementation), here's a snippet of code from another project of mine:
https://github.com/ibaned/omega_h/blob/master/src/Omega_h_mpi.h#L10
```cpp
/* on BlueGene/Q the default install
 * defines MPICH2_CONST to empty if we
 * don't define it first, causing tons
 * of compile errors.
 *
 * in addition, the mpicxx.h header is full
 * of "const MPICH2_CONST", probably as a workaround
 * for the first mistake above.
 * as a result, properly defining MPICH2_CONST to const
 * also causes compile errors.
 * luckily, we can avoid including mpicxx.h with
 * MPICH_SKIP_MPICXX.
 */
#ifdef __bgq__
#define MPICH2_CONST const
#define MPICH_SKIP_MPICXX
#endif

#include <mpi.h>
```

I think this is somewhat of a better solution

stanmoore1:
So can we put that in `kp_space_time_stack.cpp`?

stanmoore1:
Here is the compile error for reference:

```
kp_space_time_stack.cpp:397:7: error: no matching function for call to
      'MPI_Send'
      MPI_Send(s.data(), string_size, MPI_CHAR, 0, 42, MPI_COMM_WORLD);
      ^~~~~~~~
/bgsys/drivers/V1R2M4/ppc64/comm/include/mpi.h:811:5: note: candidate function
      not viable: no known conversion from 'const value_type *'
      (aka 'const char *') to 'void *' for 1st argument
int MPI_Send(MPICH2_CONST void*, int, MPI_Datatype, int, int, MPI_Comm) ...
```

stanmoore1:
@ibaned you solution does compile but generates a warning:

```
kp_space_time_stack.cpp:402:18: warning: argument type 'const value_type *'
      (aka 'const char *') doesn't match specified 'MPI' type tag that requires
      'char *' [-Wtype-safety]
      MPI_Send(s.data(), string_size, MPI_CHAR, 0, 42, MPI_COMM_WORLD);
               ~~^~~~~~               ~~~~~~~~
1 warning generated.
```

ibaned:
Whoa... they bothered to implement advanced MPI-aware warnings but didn't bother to make them const-correct? Thats pretty sad... 

It looks like the PETSc folks pointed this out three years ago and it was fixed, but probably IBM hasn't updated their MPI to include that fix:
https://trac.mpich.org/projects/mpich/ticket/2223?cversion=0&cnum_hist=1

Okay, I guess we should go ahead with the const-cast, pray that BlueGene/Q dies off soon, and not buy any more supercomputers without vendor commitment to update the software.

I'll merge your PR, thanks.

nmhamster:
@ibaned - the MPI used on our POWER systems is mainly custom install OpenMPI. We are working to get IBM's Spectrum installed and run through with all the TPLs for verification.

ibaned:
@nmhamster sorry for the confusion, we only see this issue on BlueGene/Q, where the lack of updates is a long-standing issue.

PR #35, opened 2018-04-08
Added demangling info to simple kernel timer

delpinux:
Just used cxxabi to improve kp_reader's rendering

### DIFF
--- a/src/tools/simple-kernel-timer/kp_kernel_info.h
+++ b/src/tools/simple-kernel-timer/kp_kernel_info.h
@@ -5,9 +5,9 @@
 #include <stdio.h>
 #include <sys/time.h>
 #include <cstring>
-
-
-
+#if defined(HAVE_GCC_ABI_DEMANGLE)
+#include <cxxabi.h>
+#endif
 double seconds() {
 	struct timeval now;
 	gettimeofday(&now, NULL);
@@ -26,7 +26,7 @@ class KernelPerformanceInfo {
 	public:
 		KernelPerformanceInfo(std::string kName, KernelExecutionType kernelType) :
 			kType(kernelType) {
-			
+
 			kernelName = (char*) malloc(sizeof(char) * (kName.size() + 1));
 			strcpy(kernelName, kName.c_str());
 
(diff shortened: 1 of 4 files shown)


src/tools/simple-kernel-timer/kp_kernel_info.h
```diff
@@ -5,8 +5,7 @@
 #include <stdio.h>
 #include <sys/time.h>
 #include <cstring>
-
-
+#include <cxxabi.h>
```
mhoemmen: This would likely break the Windows build, along with other compilers that may not support the GCC ABI.  You should protect this include and use of the GCC ABI with the appropriate macro.  Best practice would be to wrap `abi::__cxa_demangle` in a function that returns the original string if the GCC ABI is not available.  Here is an example:

https://github.com/trilinos/Trilinos/blob/master/packages/teuchos/core/src/Teuchos_TypeNameTraits.cpp

mhoemmen: The example doesn't show the right macro, but that shouldn't be too hard to figure out :-) .

delpinux: Indeed it would be easy to change, but since the library is only built through a simple Makefile (no CMake or configuration AFAIK), it is not clear to me how to make it in a clean way.
Should I just specify an optional macro in the Makefile and add a Readme file for this particular timer?

crtrott: Yeah just add a define in the Makefile. Make it do demangle by default. I assume the demangle function returns the original string if it couldn't do it right? Because often it can't demangle for example because of using lambdas where the names are not standardized.

delpinux:
Job done.
I rebased my changes to clean-up history for such a little change.

mhoemmen:
@delpinux Thanks!  If you feel like working more on this, you could try the `__GXX_ABI_VERSION` macro -- if it works, users wouldn't need to touch the Makefile at all.

delpinux:
@mhoemmen Ok. I'll give it a try this evening, but I am not sure if llvm supports it.
By the way, I find this simple kernel timer is very nice and I'll share any changes that I should perform.

Finally a big thank to the whole Kokkos team. This is a great work. This tools being _la cerise sur le gateau_ as we say in France :)

mhoemmen:
@delpinux See the Teuchos example.   The idea is that you write a function `demangleName`.  The function only demangles if the GCC ABI is available.  If the GCC ABI is not available, the function just returns the (mangled) input string.

PR #38, opened 2018-05-21
Space-time-stack: Adding remainder calculation

csiefer2:
Inspired by the work of @bathmatt

### DIFF
--- a/src/tools/space-time-stack/kp_space_time_stack.cpp
+++ b/src/tools/space-time-stack/kp_space_time_stack.cpp
@@ -222,7 +222,19 @@ struct StackNode {
       os << avg_runtime << " sec ";
       os << std::fixed << std::setprecision(1);
       auto percent_kokkos = (total_kokkos_runtime / total_runtime) * 100.0;
-      os << percent << "% " << percent_kokkos << "% " << imbalance << "% " << number_of_calls << " " << name;
+
+      // Sum over kids if we're a region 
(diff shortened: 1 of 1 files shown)


mhoemmen:
@csiefer2 FYI, @ibaned pointed out a difficulty over the phone today: Kokkos Profiling expects that regions always form a tree -- any two regions are either disjoint, or one fully encloses the other.  However, MueLu's use of RCP to handle TimeMonitor breaks this assumption.  For example:
```
RCP<TimeMonitor> mon = rcp (new TimeMonitor (*TimeMonitor::getNewCounter ("A")));
code_A_to_time ();
mon =  rcp (new TimeMonitor (*TimeMonitor::getNewCounter ("B")));
code_B_to_time ();
// ... etc. ...
```
The problem is that the right-hand side of `mon = rcp (new TimeMonitor (...));` gets evaluated first, and creates the second TimeMonitor, _before_ the first TimeMonitor's destructor gets called.  This means that the second TimeMonitor instance overlaps in time with the first one.  @ibaned noticed that this can result in timing differences in practice -- just a few nanoseconds, but it breaks the tree invariant and thus causes some regions to have artificially small timings.

csiefer2:
You OK to merge this then?

@mhoemmen As for the overlapping timer issue, this is due to the fact that Kokkos timers are more restrictive than the Teuchos ones, and whoever stuck the former under the latter did not check downstream packages for errors this change introduced.

PR #41, opened 2018-07-09
Remove macro definition from Makefile as suggested by @mhoemmen

delpinux:
- check for __GXX_ABI_VERSION.
- create demangleName C-like function
- report this changes to simple-kernel-timer-json

### DIFF
--- a/src/tools/simple-kernel-timer-json/kp_kernel_info.h
+++ b/src/tools/simple-kernel-timer-json/kp_kernel_info.h
@@ -6,7 +6,26 @@
 #include <sys/time.h>
 #include <cstring>
 
+#if defined(__GXX_ABI_VERSION)
+#define HAVE_GCC_ABI_DEMANGLE
+#endif
 
+#if defined(HAVE_GCC_ABI_DEMANGLE)
+#include <cxxabi.h>
+#endif // HAVE_GCC_ABI_DEMANGLE
+
+char* demangleName(char* kernelName)
+{
+#if defined(HAVE_GCC_ABI_DEMANGLE)
+	int status = -1;
+	char* demangledKernelName = abi::__cxa_demangle(kernelName, NULL, NULL, &status);
+	if (status==0) {
+		free(kernelName);
+		kernelName = demangledKernelName;
+	}
+#endif // HAVE_GCC_ABI_DEMANGLE
+	return kernelName;
+}
 
 double seconds() {
 	struct timeval now;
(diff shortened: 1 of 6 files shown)


delpinux:
Observe that I did not use value of __GXX_ABI_VERSION but only checked its definition
- tested it with g++7 and 8
- worked with clang++-3.8 to clang++-7
Obviously many improvements could be performed with these tools, but is it really relevant? I feel like keeping these tools simple might be the better.

mhoemmen:
@delpinux I can't speak for Christian, but the main thing I would want is that the tools always build and do their best to demangle.  I think you've achieved that, so thanks! :-)  I spend a lot of time reviewing code so I would be happy to continue to comment more, but it looks like you're happy with it.

DavidPoliakoff:
@delpinux, we're trying to be a little more active on this repo, thanks for your contribution.

src/tools/simple-kernel-timer/kp_kernel_info.h
```diff
@@ -5,9 +5,28 @@
 #include <stdio.h>
 #include <sys/time.h>
 #include <cstring>
+
+#if defined(__GXX_ABI_VERSION)
```
mhoemmen: The function works fine with Clang as well.  I haven't tested with other compilers.

src/tools/simple-kernel-timer-json/kp_kernel_info.h
```diff
@@ -6,7 +6,26 @@
...
 
+#if defined(__GXX_ABI_VERSION)
+#define HAVE_GCC_ABI_DEMANGLE
+#endif
 
+#if defined(HAVE_GCC_ABI_DEMANGLE)
+#include <cxxabi.h>
+#endif // HAVE_GCC_ABI_DEMANGLE
+
+char* demangleName(char* kernelName)
```
mhoemmen: The definition should really go in in a source file, not in a header file.  Otherwise, you could get multiple definition link errors, etc.

src/tools/simple-kernel-timer-json/kp_kernel_info.h
```diff
@@ -6,7 +6,26 @@
...
+{
+#if defined(HAVE_GCC_ABI_DEMANGLE)
+	int status = -1;
+	char* demangledKernelName = abi::__cxa_demangle(kernelName, NULL, NULL, &status);
+	if (status==0) {
+		free(kernelName);
```
mhoemmen: This function almost certainly will leak memory, and it won't work if `kernelName` is a constant string like `"foo"`.  Why not just return a string?
```c++
char* demangledKernelName = abi::__cxa_demangle(kernelName, nullptr, nullptr, &status);
if (demangledKernelName != nullptr) {
  std::string name (demangledKernelName);
  free (demangledKernelName);
  return name;
}
else {
  return std::string (kernelName);
}
```

PR #52, opened 2019-05-13
KP Space Time Stack: Adding kernels-per-second calculation

csiefer2:
I added some code to allow summing of the number of kernel (not region) invocations through the stack.

This allows us to calculate the total number of kernel launches per second at each*region* in the tree.

This should allow us to get realistic KPS counts in apps, especially on Cuda where it matters the mose.


csiefer2:
@crtrott @dsunder @ibaned

PR #57, opened 2019-06-04
kp_kernel_timer: Replace call to get_current_dir_name

ndellingwood:
get_current_dir_name is not portable to some MacOSX systems, leading to
errors when trying to compile the kp_kernel_timer tool like:

error: 'get_current_dir_name' was not declared in this scope

This commit replaces that call with use of getcwd


ndellingwood:
@ibaned fixed

PR #60, opened 2019-07-01
timemory-connector

jrmadsen:
@crtrott Here is the initial TiMemory connector. I wrote an extensive README and there is a sample with the output at the end.

Let me know what you think.


jrmadsen:
@DavidPoliakoff Is there a testing suite for these tools? This is working with the sample and with a tutorial but I didn't find any CI or tests.

DavidPoliakoff:
@jrmadsen , that's still an action item we need to take on. I'll probably make sure we don't accidentally explode the unit test suite, but we don't have anything formal just yet

src/tools/timemory-connector/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
+cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
+
+if("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_BINARY_DIR}")
+    message(STATUS "Error! Building from the source directory may overwrite Makefile")
+    message(STATUS "Remove 'CMakeCache.txt' and 'CMakeFiles' and build is separate directory")
```
DavidPoliakoff: Nit: `build in separate directory`

src/tools/timemory-connector/README.md
```diff
@@ -0,0 +1,404 @@
...
+- `cupti_activity`/`cupti_counters` and `gpu_roofline` types
+- Any hardware counters + `caliper` (when Caliper is also configured to collect hardware counters)
+
+```console
+export TIMEMORY_COMPONENT_LIST_INIT="cupti_activity,papi_array"
+```
+
+## Run kokkos application with PAPI recording enabled
+
+When timemory was installed with PAPI support, assuming `CMAKE_INSTALL_RPATH_USE_LINK_PATH=ON`, the timemory
```
DavidPoliakoff: This is mildly different from above, up at line 55 I need to have the PAPI location always, down here there are CMake options which will sometimes handle this for me (though it looks from the CMakeLists like you force those on in the connector itself). Double check that this and line 55 line up how you want to and resolve this thread and I'll call that good

jrmadsen: Yea so I think that was probably just too much info. I default install with the rpath encoded now but I think I didn't at the time.

I actually think I'm going to replace using the library interface with a new runtime bundler that uses the template interface so there isn't an explicit link dependency and also allow users to write their own components and plug them in if they want to.

jrmadsen:
@DavidPoliakoff Ready to go

jrmadsen:
@DavidPoliakoff I have kind of deprecated using the Makefile. It is honestly not particularly easy at this point to write a Makefile for timemory because the external library dependencies are so modular. I still have it there bc, of course, it is possible but I put in the docs that it is highly recommended to build with CMake, eventually I think I am going to generate something like a `timemory-config` or a `timemory.inc` Makefile that will get generated when I create all the CMake INTERFACE libraries (that might actually be a great CMake module distributed with the source and I think I'll run the idea by the CMake devs).

PR #66, opened 2019-10-01
Implemented the region calls in nvprof-connector

DavidPoliakoff:
The NVProf connector didn't use the push/pop region calls, this just adds that in

### DIFF
--- a/src/tools/nvprof-connector/kp_nvprof_connector.cpp
+++ b/src/tools/nvprof-connector/kp_nvprof_connector.cpp
@@ -4,12 +4,14 @@
 #include <cstdlib>
 #include <cstring>
 #include <unordered_map>
+#include <deque>
 #include <string>
 #include <iostream>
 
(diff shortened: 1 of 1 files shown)


src/tools/nvprof-connector/kp_nvprof_connector.cpp
```diff
@@ -4,12 +4,14 @@
...
 #include <cstring>
 #include <unordered_map>
+#include <deque>
 #include <string>
 #include <iostream>
 
 #include "nvToolsExt.h"
 
 static std::unordered_map<uint64_t, nvtxRangeId_t> range_map;
+static std::deque<nvtxRangeId_t> region_range_stack;
```
dalg24: If you need a stack, use a stack.

DavidPoliakoff: Done

PR #80, opened 2020-02-25
Update copyright file structure

crtrott:


DavidPoliakoff:
Looks good to me, but before merging something like this I'd like a second opinion, since we discussed this design. @dalg24, any thoughts?

dalg24:
> Looks good to me, but before merging something like this I'd like a second opinion, since we discussed this design. @dalg24, any thoughts?

As much as I dislike the current directory structure, the change will probably upset users.

Not a big fan of the argument for the `build-all.sh` script

crtrott:
I take this as a ringing endorsement @dalg24 ...

DavidPoliakoff:
@dalg24 I'll probably do a second pass at what a Kokkos Tools build system should look like at some point in the future.

I'll merge this, it seems good enough

PR #89, opened 2020-05-07
Apollo Tuner (for discussion)

DavidPoliakoff:
This is just for discussion with some collaborators on the tool


src/tools/apollo-tool/apollo-tuner.cpp
```diff
@@ -1,142 +1,179 @@
...
+  size_t variable_ids[max_variables];
+  size_t num_variables;
+};
 
 int choices[max_choices];
 
-Apollo* apollo; 
+namespace std {
+template <> struct less<variableSet> {
+  bool operator()(const variableSet &l, const variableSet &r) {
```
mhoemmen, line 38: I'm curious why you would want to specialize `std::less`, rather than just define `operator<` for your type.  Unspecialized `std::less` just calls `operator<`.

DavidPoliakoff: No conscious reason, I'm not too familiar with achieving compatibility with `std`-namespaced containers and the code I'm aping using this pattern

mhoemmen: It's both easier to define `operator<`, and more in line with developers' expectations.  If you do that, you'll get the `std::less` specialization for free.  I would find it weird to have a `std::less` specialization, but not have `operator<` work.

DavidPoliakoff: If that's what's standard I'll probably implement it when it comes time to merge. 

For now I'm not sure whether I might not just tear this all down and do something completely different, but it's a good comment if this makes it the distance. Thanks!

src/tools/apollo-tool/apollo-tuner.cpp
```diff
@@ -231,37 +253,43 @@ extern "C" void kokkosp_request_values(
...
   }
-  auto region = tuning_regions[tuningProblem];
+  auto region = tuning_regions[tuningProblem].second;
   tuned_contexts[contextId] = region;
   region->begin();
   //std::cout <<"Cont: "<<numContextVariables<<std::endl;
   for (int x = 0; x < numContextVariables; ++x) {
-    if (untunables.find(contextValues[x].id) == untunables.end()) {
+    if (untunables.find(contextValues[x].type_id) == untunables.end()) {
       region->setFeature(variableToFloat(contextValues[x]));
```
jjwilke, line 297: I don't know anything about Apollo - but - features are only allowed to be floats? That seems really limiting.

jjwilke: and dangerous

src/tools/apollo-tool/apollo-tuner.cpp
```diff
@@ -221,7 +243,7 @@ extern "C" void kokkosp_request_values(
   // "<<numTuningVariables<<" tuning\n";
   int choiceSpaceSize = 1;
```
jjwilke, line 253: It seems feasible to me that the space size could exceed a billion and overflow the int, but highly unlikely to exeed 64 bits. I would use a 64-bit here just to track overflow.

src/tools/apollo-tool/apollo-tuner.cpp
```diff
@@ -231,37 +253,43 @@ extern "C" void kokkosp_request_values(
...
   region->begin();
   //std::cout <<"Cont: "<<numContextVariables<<std::endl;
   for (int x = 0; x < numContextVariables; ++x) {
-    if (untunables.find(contextValues[x].id) == untunables.end()) {
+    if (untunables.find(contextValues[x].type_id) == untunables.end()) {
       region->setFeature(variableToFloat(contextValues[x]));
     }
   }
 
   int policyChoice = region->getPolicyIndex();
```
jjwilke, line 301: See above. I think int is fine here... unless apollo returns a larger integer.  I would probably `auto` this just in case Apollo adds support for even larger search spaces and we want to handle >1e9 search spaces.

include/impl/Kokkos_Profiling_C_Interface.h
```diff
@@ -211,13 +231,16 @@ struct Kokkos_Profiling_EventSet {
...
+  Kokkos_Profiling_endFenceFunction end_fence;
+  char profiling_padding[14 * sizeof(function_pointer)];
   Kokkos_Tools_outputTypeDeclarationFunction declare_output_type;
   Kokkos_Tools_inputTypeDeclarationFunction declare_input_type;
   Kokkos_Tools_requestValueFunction request_output_values;
+  Kokkos_Tools_contextBeginFunction begin_tuning_context;
   Kokkos_Tools_contextEndFunction end_tuning_context;
   Kokkos_Tools_optimizationGoalDeclarationFunction declare_optimization_goal;
-  char padding[235 *
+  char padding[234 *
```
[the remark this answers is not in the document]
DavidPoliakoff: This is copied over from the main Kokkos repo, we should file an issue there, but you should file that issue, that code is kinda trash

include/impl/Kokkos_Profiling_C_Interface.h
```diff
@@ -211,13 +231,16 @@ struct Kokkos_Profiling_EventSet {
   Kokkos_Profiling_profileEventFunction profile_event;
   Kokkos_Profiling_beginDeepCopyFunction begin_deep_copy;
   Kokkos_Profiling_endDeepCopyFunction end_deep_copy;
-  char profiling_padding[16 * sizeof(function_pointer)];
+  Kokkos_Profiling_beginFenceFunction begin_fence;
+  Kokkos_Profiling_endFenceFunction end_fence;
+  char profiling_padding[14 * sizeof(function_pointer)];
```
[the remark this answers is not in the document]
DavidPoliakoff: Same, this is a mainline Kokkos problem

include/impl/Kokkos_Profiling_C_Interface.h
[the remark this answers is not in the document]
DavidPoliakoff: Main Kokkos

include/impl/Kokkos_Profiling_C_Interface.h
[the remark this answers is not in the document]
DavidPoliakoff: Main Kokkos

include/impl/Kokkos_Profiling_C_Interface.h
```diff
@@ -172,24 +190,26 @@ struct Kokkos_Tools_VariableInfo {
...
+    const size_t count, struct Kokkos_Tools_VariableValue*);
+typedef void (*Kokkos_Tools_contextBeginFunction)(const size_t);
+typedef void (*Kokkos_Tools_contextEndFunction)(
+    const size_t, struct Kokkos_Tools_VariableValue);
 typedef void (*Kokkos_Tools_optimizationGoalDeclarationFunction)(
-    const size_t, const Kokkos_Tools_OptimzationGoal& goal);
+    const size_t, const struct Kokkos_Tools_OptimzationGoal goal);
 
-using function_pointer = void (*)();
+typedef void (*function_pointer)();
```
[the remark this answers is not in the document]
DavidPoliakoff: Main Kokkos

src/tools/apollo-tool/apollo-tuner.cpp
DavidPoliakoff: That is the case, yes

src/tools/apollo-tool/apollo-tuner.cpp
DavidPoliakoff: Limitation of Apollo

src/tools/apollo-tool/apollo-tuner.cpp
```diff
@@ -231,37 +253,43 @@ extern "C" void kokkosp_request_values(
...
 
   for (int x = 0; x < numTuningVariables; ++x) {
-    if (unlearnables.find(tuningValues[x].id) == unlearnables.end()) {
+    if (unlearnables.find(tuningValues[x].type_id) == unlearnables.end()) {
       int set_size = tuningValues[x].metadata->candidates.set.size;
       int local_policy_choice = policyChoice % set_size;
       tuningValues[x] = mvv(local_policy_choice, tuningValues[x], tuningValues[x].metadata->candidates.set);
```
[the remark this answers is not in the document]
DavidPoliakoff: The contract of Tuning says that the values provided to the tool may not crash the program. If this "learn" function is implemented as `return;` the code shouldn't crash (this also saves the application in instances where they *don't* load a tool). 

Unlearnables are filled in at line 180, but that was before an interface change, nothing is currently unlearnable anymore, I'll remove that

src/tools/apollo-tool/apollo-tuner.cpp
DavidPoliakoff: I like that. Future-proof us if Apollo has that ability. I'm not sure OpenCV (which Apollo uses) allows that to be an int64, but if they do, that's a win

src/tools/apollo-tool/apollo-tuner.cpp
jjwilke: Yeah, more a question of if you overflow `max_choices`. If you made it a 64-bit int temporarily - even if it's not allowed to exceed 32-bits - you would be more robust detecting overflow.  Unlikely to occur, obviously.

src/tools/apollo-tool/apollo-tuner.cpp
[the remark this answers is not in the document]
jjwilke: Ah, okay, was curious what the contract of `kokkosp_request_values` was.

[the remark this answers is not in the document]
DavidPoliakoff: It would probably be easier for readers if I'd actually, you know, written that down somewhere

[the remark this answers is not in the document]
DavidPoliakoff: Yikes, thanks

src/tools/apollo-tool/apollo-tuner.cpp
```diff
@@ -187,27 +191,45 @@ float variableToFloat(Kokkos::Tools::Experimental::VariableValue value) {
...
     value.value.int_value = set.values.int_value[index];  
   break;
-  case Kokkos::Tools::Experimental::ValueType::kokkos_value_floating_point:
+  case Kokkos::Tools::Experimental::ValueType::kokkos_value_double:
     value.value.double_value = set.values.double_value[index];  
   break;
-  case Kokkos::Tools::Experimental::ValueType::kokkos_value_text:
-    value.value.string_value = set.values.string_value[index];  
+  case Kokkos::Tools::Experimental::ValueType::kokkos_value_string:
+    strncpy(value.value.string_value,set.values.string_value[index],65);
```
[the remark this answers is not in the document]
DavidPoliakoff: This should probably be handled Kokkos-side, this'll blow up every tool :(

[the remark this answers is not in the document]
DavidPoliakoff: I handled the tool-side of it though

src/tools/apollo-tool/apollo-tuner.cpp
```diff
@@ -231,37 +253,43 @@ extern "C" void kokkosp_request_values(
...
-    std::string name = "debug_name_"+std::to_string(created_regions);
-    auto region =
-        new Apollo::Region(numContextVariables, name.c_str(), choiceSpaceSize);
-    tuning_regions[tuningProblem] = region;
+    std::string name = get_region_name(tuningProblem);
+    Apollo::Region* region;
+    if(file_exists(name)) {
+      region = new Apollo::Region(numContextVariables,name.c_str(), choiceSpaceSize, name); 
+    }
+    else{
```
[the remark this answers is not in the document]
DavidPoliakoff: I grabbed the iterator. I know logically it's good to avoid the double lookup, but I hate how this code now looks

src/tools/apollo-tool/apollo-tuner.cpp
```diff
@@ -0,0 +1,313 @@
...
+  for (int x = 0; x < numContextVariables; ++x) {
+    if (untunables.find(contextValues[x].type_id) == untunables.end()) {
+      tuningProblem.variable_ids[numValidVariables++] = contextValues[x];
+    }
+  }
+  for (int x = 0; x < numTuningVariables; ++x) {
+      tuningProblem.variable_ids[numValidVariables++] = tuningValues[x];
+  }
+  tuningProblem.num_variables = numValidVariables;
+  auto iter = tuning_regions.emplace(tuningProblem, std::make_pair("",nullptr));
```
jjwilke, line 262: I see you don't like this code!

Is this more aesthetically pleasing?
````cpp
auto iter = tuning_regions.find(tuning_problem);
if (iter == tuning_regions.end()){
  //bla bla bla
  iter = tuning_regions.emplace(...)
}
region = iter->second.second;
````

DavidPoliakoff: Yeah, I'm not a huge fan of either, but I can't even pretend it's rational. I appreciate the suggestion, but I think what we have is decent, and appears to match standard practice

DavidPoliakoff:
Thanks for the efforts, reviewers. I'm actually moving this to llnl/apollo#12, if they'll merge it. Rather than have a Kokkos connector library to Apollo, just include the Kokkos functionality in Apollo, save our users some steps

PR #91, opened 2020-06-03
Labrador Tuning Toolkit

DavidPoliakoff:
Initial, crazy, unpolished commit of the Labrador Tuning Toolkit. Most of the diff comes from the bundled sqlite3 source.

The goal of this toolkit is to enable tuning while being as simple for build and integration as possible. Its model is an offline one. A user loads `liblabrador-explorer.so` into as many runs as they want, and the tool will cycle through policies recording data. This outputs a database that they analyze using `labrador-scholar.py`, the output of which is a C++ program that can be piped to `labrador-retriever.cpp`, which can then be built into a tool (by default called `liblabrador-retriever.so`, which shouldn't be changed under any circumstances). This tool can then tune runs of the application like those it has seen before.

The dependencies of the toolkit are bundled with it, all you need is a C++11 compliant compiler and Python 3.5 or greater.

Note that these things are separable, somebody literate in machine learning could write something a lot better than my brute force tuner that replaces `labrador-scholar.py`, though I'd ask they have the decency to stick to the naming scheme


vlkale:
Identifying a mixture / hybridization of policies may benefit,  e.g., Mixed static/dynamic loop scheduling strategy/policy, as shown [here](https://dl.acm.org/doi/10.1145/2642769.2642788).

A machine learning approach in Labrador-scholar.py would identify the best aspect fraction of dynamic iterations in the mixed static/dynamic loop scheduling strategy. Note that the mixed static/dynamic scheduling policy has been implemented in RAJA fork here.

DavidPoliakoff:
Hey Vivek, just so you know this PR is pretty dead, I'm not at the labs anymore. If you're looking to promote this with Kokkos folks, recommend reaching out to @crtrott or @dalg24 , and naming the tool with a canine theme

vlkale:
@DavidPoliakoff Got it. I am working with Christian and Damien to move forward work on Kokkos tools, with Kokkos library parameter application parameter performance tuning being a particular interest of mine. I will use a canine name :)

vlkale:
> @DavidPoliakoff Got it. I am working with Christian and Damien to move forward work on Kokkos tools, with Kokkos library parameter application parameter performance tuning being a particular interest of mine. I will use a canine name :)
> 

Closing this PR as per discussion with @DavidPoliakoff and @khuck  this past week. This will be integrated into efforts for the Kokkos Tools connector for Apex auto-tuning capabilities.

PR #98, opened 2020-08-13
[WIP] Kokkos View core-dump utility tool

DavidPoliakoff:
Intended to share with a code team, not yet ready for review


keitat:
Just a quick question. My team is working on resilient execution space and memory space.  Is it related to this work?

DavidPoliakoff:
Keita, more later, but the short version (please excuse my terseness):

This tool allows you to specify a kernel. Each time that kernel runs, every
live Kokkos allocation gets copied back to the host.

On a segfault/abort/error: this tool dumps out all the Views to a file

I don't think it's related, though I'd love to know more about what your
team is doing to tell you for sure :)

crtrott:
This is not really related. While it uses similar effective technologies, it is aimed at debugging not resilience at all. It is more like a core-dump utility and I doubt anyone would consider core-dumps a resilience mechanism.

PR #99, opened 2020-08-25
Update nvprof-connector to use nsys-compatible interface

DavidPoliakoff:
The `nvtxRangeStartA` interface doesn't lead to names appearing in the nsys GUI, the RangePush/RangePop interface does. Note that I left the pushRegion/popRegion as is in the (Kokkos) push_region/pop_region call, as these aren't involved in naming kernels


DavidPoliakoff:
@dalg24 I'll try to handle that in bulk (a PR to make the tools match the Kokkos coding standards)

PR #104, opened 2020-09-01
Develop -> Master

DavidPoliakoff:
About five people have needed the fix we did to the nvprof-connector, merging so they have that fix in the main branch


dalg24 (approved):
Do we still need two branches?

PR #110, opened 2020-11-16
Fixup SimpleKernelTimer JSON format output

dalg24:


DavidPoliakoff (approved):
Oof. Thanks for the fix

PR #117, opened 2021-03-03
Sampling overlay prototype

DavidPoliakoff:
Based on kokkos/kokkos#3812

One hope I have for that design is that our simple tools can be given complex capabilities using "overlays," (a concept I'm shamelessly stealing from @nmhamster  and @jrmadsen ). This PR introduces some infrastructure to enable that concept, and two example overlays.

First, I formalized the process for forwarding your callbacks to a next tool, using a "callback_chainer" struct, which handles loading a next library, grabbing callbacks, all that mess.

The first example overlay is a trivial "passthrough" overlay. All it does is forward events, I basically used it to prove out the callback_chainer thing.

The other is a sampler overlay. This turns off Kokkos global fencing, and then every 101 kernel invocations (that should be a settable parameter) it forwards the kernel to an underlying tool. Fences are only introduced in the case where the kernel is profiled. I've tested this with the space-time-stack and TAU (TAU through modifying tau_exec).

What's neat here is that space-time-stack is completely unmodified, but gets this sampling capability "for free." It's kind of a deeply weird version of a modular tools framework.

Not sure how I feel about this design as a thing to merge, mainly just putting it up as another proof of concept for the Core PR

### DIFF
--- a/common/sampler/sampler.cpp
+++ b/common/sampler/sampler.cpp
@@ -1,0 +1,186 @@
+#include <cinttypes>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+#include <cxxabi.h>
+#include <dlfcn.h>
+#include <impl/Kokkos_Profiling_Interface.hpp>
+#include <iostream>
+#include <regex>
+#include <string>
+#include <tool_chain.hpp>
+#include <unordered_set>
+#include <vector>
+static Kokkos::Tools::callback_chainer next;
+using action_type = Kokkos::Tools::Experimental::ToolActions;
+action_type actions;
+static uint64_t sample_rate = 101;
+
+extern "C" void kokkosp_begin_parallel_scan(const char *name,
+                                            const uint32_t devID,
+                                            uint64_t *kID) {
+  *kID = 0;
+  static uint64_t encounter;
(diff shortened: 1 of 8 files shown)


common/sampler/sampler.cpp
```diff
@@ -0,0 +1,186 @@
...
+static uint64_t sample_rate = 101;
+
+extern "C" void kokkosp_begin_parallel_scan(const char *name,
+                                            const uint32_t devID,
+                                            uint64_t *kID) {
+  *kID = 0;
+  static uint64_t encounter;
+  ++encounter;
+  if ((encounter % sample_rate) == 0) {
+    actions.fence(0);
```
vlkale, line 26: Making a comment for myself (David, you don't need to respond to this of course) that I think this parameter to fence function should not have to be zero.

vlkale:
I am closing this PR as no further work needs to be done on it currently within the Kokkos team, and its underlying goals have been accomplished via the recently merged #213. There are some new ideas here not in PR #213. However, having talked with @(DavidPoliakoff), this can be brought back to life if there is a Kokkos Tools user request / Kokkos Tools GitHub issue that necessitates it.

PR #119, opened 2021-04-12
Allow HIP and SYCL in space-time-stack

masterleinad:
With this patch information for allocations in `Host`, `Cuda`, `HIP`, and `SYCL` memory spaces are recorded in `space-time-stack`.

### DIFF
--- a/profiling/space-time-stack/kp_space_time_stack.cpp
+++ b/profiling/space-time-stack/kp_space_time_stack.cpp
@@ -85,14 +85,14 @@ enum Space {
 enum { NSPACES = 4 };
 
 Space get_space(SpaceHandle const& handle) {
-  if (strcmp(handle.name, "Cuda") == 0 || strcmp(handle.name, "CudaUVM") == 0 ||
-      strcmp(handle.name, "CudaHostPinned")==0 )
+  // check that name starts with "Cuda"
+  if (strncmp(handle.name, "Cuda", 4) == 0)
     return SPACE_CUDA;
-  if (strcmp(handle.name, "SYCLDeviceUSM") == 0 ||
-      strcmp(handle.name, "SYCLSharedUSM") == 0)
+  // check that name starts with "SYCL"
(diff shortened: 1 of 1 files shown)


profiling/space-time-stack/kp_space_time_stack.cpp
```diff
@@ -76,16 +77,26 @@ struct SpaceHandle {
...
-enum { NSPACES = 2 };
+enum { NSPACES = 4 };
 
 Space get_space(SpaceHandle const& handle) {
-  switch (handle.name[0]) {
-    case 'H': return SPACE_HOST;
-    case 'C': return SPACE_CUDA;
-  }
+  if (strcmp(handle.name, "Cuda") == 0 || strcmp(handle.name, "CudaUVM") == 0 ||
+      strcmp(handle.name, "CudaHostPinned")==0 )
```
dalg24, line 89:
```suggestion
  if (strncmp(handle.name, "Cuda", 4) == 0)  // check that name starts with "Cuda"
```

dalg24 (approved):
I would use `strncmp` but not blocking

DavidPoliakoff (changes requested):
Agree with Damien's comment, but I'll block on it. This way if we do something weird to the Cuda backend, we still should see this work

profiling/space-time-stack/kp_space_time_stack.cpp
masterleinad: Changed for `CUDA`, `HIP` and `SYCL`.

masterleinad:
@DavidPoliakoff Is there anything else you want me to do here?

DavidPoliakoff (approved):

DavidPoliakoff:
@masterleinad: in the long term, we're phasing out develop, but for the moment could you retarget against develop?

masterleinad:
Sure, targeting `develop` now.

DavidPoliakoff:
I don't know how I forgot to merge this, apologies

PR #125, opened 2021-07-28
Use NVTX range push/pop instead of start/end in push/pop_profile_region

maxpkatz:
The current profiling APIs mix nvtxRangePush/Pop and nvtxRangeStart/End. The consequence is that Kokkos View construction and parallel for operations may use a different NVTX API than what the user does with explicit profiling regions in their code, which can be confusing to look at in an Nsight Systems timeline. If everything uses nvtxRangePush() and nvtxRangePop() then all regions should be nicely nested on the timeline. Also, nvtxRangePush() and nvtxRangePop() effectively manage their own stack, so there is no need for kokkos-tools to do this.

### DIFF
--- a/profiling/nvprof-connector/kp_nvprof_connector.cpp
+++ b/profiling/nvprof-connector/kp_nvprof_connector.cpp
@@ -51,8 +51,6 @@
 
 #include "nvToolsExt.h"
 
-static std::unordered_map<uint64_t, nvtxRangeId_t> range_map;
-static std::stack<nvtxRangeId_t> region_range_stack;
 static uint64_t nextKernelID;
 
 extern "C" void kokkosp_init_library(const int loadSeq,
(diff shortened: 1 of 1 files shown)


maxpkatz:
Here is an example from Parthenon where the parallel for operation DepositParticles is a subset of the function DepositParticles (which is marked up with pushRegion("Task_Particles_DepositParticles")), which is called from within Driver_Main. Because of the inconsistency in the push/pop versus start/end APIs, we don't see the logical nesting of operations the user might expect.

before

But by making everything use nvtxRangePush and nvtxRangePop, the logical nesting that is desired is what appears.

after

dalg24:
Hey Max.  Please change the destination branch from your PR from `master` to `develop`

maxpkatz:
Done. Also cc'ing @DavidPoliakoff on this because he mentioned he might have thoughts here.

DavidPoliakoff (approved):
I'll approve, I can't think of a reason this was a problem. If we hit problems, we can revert

crtrott:
Yeah this is correct. Note we also have "sections" which are not exposed in the tool right now, which are allowed to overlap and not nest.

PR #131, opened 2021-10-30
CMake build system

### LINKED ISSUE #128: NGA FY22:  improve build system
- [x] Develop CMake build system to build all the KokkosTools
- [x] Extent the tools cmake build system to optionally build Caliper and Apex (i.e. translate typical Kokkos
options such as Kokkos_ENABLE_CUDA to the respective settings for Caliper and Apex)
- [x] Add option to build all tools as a single library, and add an initialization interface to that library, which uses
the existing KokkosEventSet capability to load a specific tool in the library upon being called with the
corresponding option (e.g. KokkosTools::initialize("space_time_stack"));

mzuzek:
Fixes #128

## Summary

With this PR we enable building the tools with CMake. Caliper and Apex are connected as git submodules and can be built from source if no existing installation is found. In build, tools try to find Kokkos and propagate it's settings (e.g. enable CUDA support) to tools and TPLs. "Single library intefrace" is introduced which allows simple linking of all enabled tools and TPLs to client application.

### Build: sample app CMake list file

```cmake
find_package(Kokkos REQUIRED)
add_directory(kokkos-tools) # TODO: CMake install exports

add_executable(myapp source.cpp)

target_link_libraries(myapp Kokkos::kokkos kokkostools) # link all tools
```

### Usage: sample app source code

```cpp
#include <Kokkos_Core.hpp>
#include "kp_all.hpp"

int main(int argc, char *argv[])
{
  const char *profiler_name = "caliper";
  const char *profiler_config = "runtime-report(profile.kokkos)";

  auto eventSet = KokkosTools::get_event_set(profiler_name, profiler_config);
  Kokkos::Tools::Experimental::set_callbacks(eventSet);

  // ... run app ... //
  Kokkos::finalize();
  return 0;
}
```

## Contents / Status

* General:
  - [x] CMake scripts to build tools;
  - [x] Single library interface linking all tools and TPLs;
  - [x] Fetch basic Kokkos settings in alternative scenarios:
    - [x] 1. from existing installation (`find_package(Kokkos)` based on `Kokkos_ROOT`);
    - [x] 2. from source if Kokkos is being built within same parent CMake project;
    - [x] 3. skip option propagation if Kokkos is not available;
  - [x] MPI support;
  - [x] Basic Windows (MSVC) build support - with Caliper and VTune;
  - [x] CMake exports for tools Installation;
  - [ ] Allow building each tool separately;
* Building tools (for traditional runtime shared library connection):
  - [x] Native tools: `space-time-stack`, `chrome-tracing`, `simple-kernel-timer`, `simple-kernel-timer-json`, `memory-hwm-mpi`, `memory-usage`, `memory-events`, `memory-hwm` and `memory-hwm-mpi`;
  - [x] Native utilities: `kernel-filter` and `kernel-logger`;
  - [x] TPL connectors:  [VTune](https://www.intel.com/content/www/us/en/develop/documentation/vtune-help/top.html), [SystemTap](https://sourceware.org/systemtap/), [Variorum](https://github.com/llnl/variorum) and [PAPI](https://bitbucket.org/icl/papi/wiki/Home);
  - [x] TPLs: [Caliper](https://github.com/LLNL/Caliper) and [Apex](https://github.com/UO-OACISS/apex) built from source (submodules) if no existing installation is detected;
  - [ ] TPLs: [apollo (`builtin-ml-library` branch)](https://github.com/llnl/apollo/tree/builtin-ml-library);
  - [ ] GPU profiler connectors:
    - [x] [_nvprof_](https://docs.nvidia.com/cuda/profiler-users-guide/index.html#nvprof-overview) (CUDA);
    - [ ] [_roctx_](https://rocmdocs.amd.com/en/latest/Installation_Guide/Tools-Installation.html#roctracer) (HIP);
* Tools available via `KokkosTools::get_event_set(profiler_name, profiler_config)` single library interface:
  - [x] All native tools;
  - [x] TPL connectors: VTune, SystemTap and Variorum;
  - [x] Caliper with configuration string support, requires WIP https://github.com/LLNL/Caliper/pull/402 branch;
  - [ ] Apex, PAPI connector and GPU connectors;
  - [ ] Native utilities (`kernel-filter` and `kernel-logger`) + implement `std::vector`/array overload of `Kokkos::Tools::Experimental::set_callbacks()` for registering multiple event sets ?
* Documentation and examples:
  - [x] Add [example](https://github.com/NexGenAnalytics/kokkos-tools/tree/cmake-build-system/example) demonstrating single library interface;
  - [ ] Update [ReadMe.md](https://github.com/kokkos/kokkos-tools/blob/master/README.md) with instructions on how to configure and build tools with CMake;
* Other / fixes:
  - [ ] Implement own tool to fetch parent path and replace `cmake_path(GET ... PARENT_PATH ...)` which requires CMake v3.20 ?

#### Future considerations:
- connect [_timemory_](https://timemory.readthedocs.io/en/develop/) (like _Caliper_ and _Apex_);
- consider introducing lazy profiler initialization (`kokkosp_init_library()` called after `Kokkos::initialize()`);


CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
+cmake_minimum_required(VERSION 3.16 FATAL_ERROR)
+
+project(KokkosTools CXX)
+
+# Set policies
+cmake_policy(SET CMP0054 NEW) # simplify if()
+cmake_policy(SET CMP0057 NEW) # support IN_LIST in if()
```
dalg24: These two are pointless if we require cmake 3.16

CMakeLists.txt
```diff
@@ -0,0 +1,169 @@
...
+
+# Tools settings
+set(KOKKOSTOOLS_ENABLE_CALIPER ON CACHE BOOL "Enable building Caliper library")
+set(KOKKOSTOOLS_ENABLE_APEX    ON CACHE BOOL "Enable building Apex library")
+
+# Check Kokkos settings
+if(TARGET Kokkos::kokkos)
+  message(STATUS "Using Kokkos settings for KokkosTools (source: ${Kokkos_SOURCE_DIR})")
+else()
+  message(WARNING "Not using Kokkos for tools configuration\n\tNote: using settings from installed is Kokkos not yet supported")
```
DavidPoliakoff, line 29: Oh, this is a really slick idea

DavidPoliakoff:
I like where you're starting with this. Also looping in @daboehme in case he has opinions about how we're playing with Caliper here

CMakeLists.txt
```diff
@@ -0,0 +1,169 @@
...
+
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
+
+message(STATUS)
+message(STATUS Configuring Kokkos-Tools)
+message(STATUS)
+
+# Tools settings
+set(KOKKOSTOOLS_ENABLE_CALIPER ON CACHE BOOL "Enable building Caliper library")
+set(KOKKOSTOOLS_ENABLE_APEX    ON CACHE BOOL "Enable building Apex library")
```
jrmadsen, line 23:
```suggestion
option(KOKKOSTOOLS_ENABLE_CALIPER "Enable building Caliper library" ON)
option(KOKKOSTOOLS_ENABLE_APEX    "Enable building Apex library"    ON)
```

CMakeLists.txt
jrmadsen: Why would there be a `Kokkos::Kokkos` target w/o a `find_package(Kokkos)`?

CMakeLists.txt
```diff
@@ -0,0 +1,169 @@
...
+#       * error out otherwise: does it make sense to look into Kokkos
+#         source code without knowing it's *actual* build options ?
+
+#Turn MPI support off: (arbitrary or if Kokkos has MPI disabled)
+#CFLAGS += -DUSE_MPI=0
+
+# CXXFLAGS=-O3 -std=c++11 -g -shared -fPIC
+# -Wall -Wextra
+
+set(CMAKE_CXX_COMPILER ${Kokkos_CXX_COMPILER})
```
jrmadsen, line 44: I _really_ wouldn't recommend doing this. Emit an `AUTHOR_WARNING` (so it can be suppressed) but don't force the same compiler bc you might end up with tools which have bugs w/ certain compilers/compiler versions. IIRC, the whole reason @DavidPoliakoff was so strict about KokkosTools having a C interface was avoid issues w/ KokkosTools being compiled with a different C++ compiler as kokkos.

CMakeLists.txt
```diff
@@ -0,0 +1,169 @@
...
+set(CMAKE_CXX_STANDARD ${CMAKE_CXX_STANDARD_DEFAULT})
```
jrmadsen, line 45: Cache this so it can be changed. You cannot override a local variable with a `-DCMAKE_CXX_STANDARD=...` on the command line.

CMakeLists.txt
```diff
@@ -0,0 +1,169 @@
...
+
+# Libraries
+find_package(PAPI)
```
jrmadsen, line 48: There should be an option to enable/disable papi support

CMakeLists.txt
```diff
@@ -0,0 +1,169 @@
...
+if(PAPI_FOUND)
+  cmake_path(GET PAPI_INCLUDE_DIR PARENT_PATH PAPI_ROOT)
+endif()
+find_package(OpenMP)
+find_package(MPI)
```
jrmadsen, line 53: There should be an option to enable/disable MPI support

CMakeLists.txt
```diff
@@ -0,0 +1,169 @@
...
+
+set(CMAKE_CXX_COMPILER ${Kokkos_CXX_COMPILER})
+set(CMAKE_CXX_STANDARD ${CMAKE_CXX_STANDARD_DEFAULT})
+
+# Libraries
+find_package(PAPI)
+if(PAPI_FOUND)
+  cmake_path(GET PAPI_INCLUDE_DIR PARENT_PATH PAPI_ROOT)
+endif()
+find_package(OpenMP)
```
jrmadsen, line 52: I'm not sure where OpenMP is needed but there should be an option to enable/disable OpenMP support.

CMakeLists.txt
jrmadsen: This should (IMO) take the form of:

```cmake
option(KOKKOSTOOLS_USE_PAPI "Enable PAPI support" ON)
if(KOKKOSTOOLS_USE_PAPI)
    find_package(PAPI REQUIRED)
endif()
```

So I used to do what you did here in the past. Experience has taught me that it is _much_ better to be explicit: make it an option and fail if it is not found. Especially since you have Caliper and Apex subprojects -- users will likely miss that PAPI was not found bc there will be so many messages from adding those subprojects.

CMakeLists.txt
```diff
@@ -0,0 +1,169 @@
...
+
+# Note: Caliper docs https://software.llnl.gov/Caliper/build.html
+#       might miss some options, check the top CMakeLists.txt:
+#       https://github.com/LLNL/Caliper/blob/master/CMakeLists.txt
+macro(configure_caliper)
+  #set(BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS}) # Build shared (ON - default) or static library (OFF).
+  #set(BUILD_DOCS OFF)   # Build documentation.
+  #set(BUILD_TESTING OFF) # Build unit tests.
+  #set(WITH_FORTRAN OFF) # Build and install Fortran wrappers.
+  set(WITH_KOKKOS ON) # Enable Kokkos profiling support
```
jrmadsen, line 110: So I added support in Caliper to do:

```cmake
set(CALIPER_OPTION_PREFIX ON CACHE INTERNAL "Prefix caliper options with CALIPER_" FORCE)
```

and all the caliper options, e.g. `WITH_KOKKOS` appear in the cache as `CALIPER_WITH_KOKKOS`. This makes it very clear what are caliper options when it is a subproject.

CMakeLists.txt
```diff
@@ -0,0 +1,169 @@
...
+
+  # Build with support for the OpenMP tools interface.
+  if(OpenMP_FOUND)
+    set(WITH_OMPT ON)
+  else()
+    set(WITH_OMPT OFF)
+  endif()
```
jrmadsen, line 117:
```suggestion
```

Let the user specify `CALIPER_WITH_OMPT=ON`, there is no way to override this local variable if there is a problem with the Caliper OMPT build

CMakeLists.txt
```diff
@@ -0,0 +1,169 @@
...
+    set(PAPI_PREFIX ${PAPI_ROOT})
+  endif()
+  # Build adapters to forward Caliper annotations to Intel’s VTune annotation API.
+  if(VTune_ROOT)
+    set(WITH_VTUNE ON)
+    set(ITT_PREFIX ${VTune_ROOT}/sdk)
+    # TODO: Caliper doesn't seek ITT in ${VTune_ROOT}/sdk/lib64 or ${VTune_ROOT}/sdk/lib32 (only in .../lib)
+  else()
+    set(WITH_VTUNE OFF)
+  endif()
```
jrmadsen, line 138: Again, you are forcing a configuration which the user may not want. Let them specify it themselves or, at the very least, set these as cache variables so there are default settings but the user can change them.

CMakeLists.txt
```diff
@@ -0,0 +1,169 @@
...
+    set(APEX_WITH_OMPT ON)
+    # TODO: set(OMPT_ROOT ...) ?
+  else()
+    set(APEX_WITH_OMPT OFF)
+  endif()
+  # TODO: Apex tries to fetch TPLs and install them in /usr/lib
+  #       (normally failing) unless we disable them...
+  set(APEX_WITH_ACTIVEHARMONY OFF)
+  set(APEX_WITH_OTF2 OFF)
+  set(APEX_WITH_MPI OFF)
```
jrmadsen, line 158: Same as before, cache variables so the user can configure them as needed.

CMakeLists.txt
```diff
@@ -0,0 +1,169 @@
...
+endmacro()
+
+# Build TPLs
+if(KOKKOSTOOLS_ENABLE_CALIPER)
+  configure_caliper()
```
jrmadsen, line 163: I don't really see the need for a macro here. Personally, i think you should just put setting the initial cache values in this if block.

CMakeLists.txt
```diff
@@ -0,0 +1,169 @@
...
+  add_subdirectory(tpls/Caliper)
+endif()
+if(KOKKOSTOOLS_ENABLE_APEX)
+  configure_apex()
```
jrmadsen, line 167: Same comment about using a macro here

cmake/FindPAPI.cmake
```diff
@@ -0,0 +1,101 @@
...
+
+``PAPI_VERSION_STRING``
+  The version of ``papi`` found.
+
+This module defines ``PAPI::libpapi`` target for PAPI library.
+
+#]=======================================================================]
+
+# Look for the header file.
+if(NOT PAPI_INCLUDE_DIR)
```
jrmadsen, line 34:
```suggestion
```

You don't need to protect this.

cmake/FindPAPI.cmake
```diff
@@ -0,0 +1,101 @@
...
+  TARGET PAPI::libpapi APPEND PROPERTY
+  IMPORTED_CONFIGURATIONS RELEASE DEBUG)
+
+set_target_properties(
+  PAPI::libpapi PROPERTIES
+  INTERFACE_INCLUDE_DIRECTORIES "${PAPI_INCLUDE_DIR}"
+  IMPORTED_LINK_INTERFACE_LANGUAGES "C"
+  IMPORTED_LOCATION "${PAPI_LIBRARY}"
+  IMPORTED_LOCATION_RELEASE "${PAPI_LIBRARY}"
+  IMPORTED_LOCATION_DEBUG "${PAPI_LIBRARY}")
```
jrmadsen, line 101:
```suggestion
# Set up imported target
add_library(PAPI::PAPI INTERFACE IMPORTED)

target_include_directories(PAPI::PAPI INTERFACE ${PAPI_INCLUDE_DIR})
target_link_libraries(PAPI::PAPI INTERFACE ${PAPI_LIBRARY})

set(PAPI_INCLUDE_DIRS ${PAPI_INCLUDE_DIR})
set(PAPI_LIBRARIES ${PAPI_LIBRARY})
```

This is how I typically do this.

cmake/FindPAPI.cmake
```diff
@@ -0,0 +1,101 @@
...
+    PAPI_VERSION_STRING
+    "${PAPI_VERSION_STRING}")
+  string(REPLACE "," "."  PAPI_VERSION_STRING "${PAPI_VERSION_STRING}")
+endif()
+
+find_package_handle_standard_args(
+  PAPI
+  REQUIRED_VARS PAPI_LIBRARY PAPI_INCLUDE_DIR
+  VERSION_VAR PAPI_VERSION_STRING
+  HANDLE_COMPONENTS)
```
jrmadsen, line 71:
```suggestion
include(FindPackageHandleStandardArgs)
find_package_handle_standard_args(
  PAPI
  REQUIRED_VARS PAPI_LIBRARY PAPI_INCLUDE_DIR
  VERSION_VAR PAPI_VERSION_STRING)
```

You do not support components so `HANDLE_COMPONENTS` is unnecessary.

cmake/FindPAPI.cmake
```diff
@@ -0,0 +1,101 @@
...
+
+# Handle missing library
+if(NOT PAPI_FOUND)
+  if(PAPI_FIND_REQUIRED)
+    message(FATAL_ERROR "PAPI library not found")
+  elseif(NOT PAPI_FIND_QUIETLY)
+    message(WARNING "PAPI library not found")
+  endif()
+  return()
+endif()
```
jrmadsen, line 81:
```suggestion
```

This is handled by `find_package_handle_standard_args`

cmake/FindPAPI.cmake
```diff
@@ -0,0 +1,101 @@
...
+#]=======================================================================]
+
+# Look for the header file.
+if(NOT PAPI_INCLUDE_DIR)
+  find_path(
+    PAPI_INCLUDE_DIR
+    NAMES papi.h
+    HINTS /usr/include /usr/local/include)
+  mark_as_advanced(PAPI_INCLUDE_DIR)
+endif()
```
jrmadsen, line 40:
```suggestion
```

cmake/FindPAPI.cmake
```diff
@@ -0,0 +1,101 @@
...
+
+if(NOT PAPI_LIBRARY)
```
jrmadsen, line 42:
```suggestion
```

cmake/FindPAPI.cmake
```diff
@@ -0,0 +1,101 @@
...
+  # Look for the library (sorted from most current/relevant entry to least).
+  find_library(
+    PAPI_LIBRARY
+    NAMES papi
+    HINTS /usr/lib /usr/local/lib)
+  mark_as_advanced(PAPI_LIBRARY)
+endif()
```
jrmadsen, line 49:
```suggestion
```

profiling/chrome-tracing/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+find_package(MPI REQUIRED)
+
+add_library(kp_chrome_tracing SHARED kp_chrome_tracing.cpp)
+
+if(NOT DEFINED USE_MPI OR USE_MPI)
```
jrmadsen, line 5:
```suggestion
if(KOKKOSTOOLS_USE_MPI AND TARGET MPI::MPI_CXX)
```

jrmadsen (commented):
Just my thoughts

CMakeLists.txt
DavidPoliakoff: In my app's CMakeLists.txt

```
add_subdirectory(kokkos) # defines the target
add_subdirectory(kokkos-tools)
```

CMakeLists.txt
```diff
@@ -0,0 +1,169 @@
...
+cmake_policy(SET CMP0054 NEW) # simplify if()
+cmake_policy(SET CMP0057 NEW) # support IN_LIST in if()
+cmake_policy(SET CMP0111 NEW) # error if library not found
+
+# Disable in-source builds to prevent source tree corruption.
+if(CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR)
+  message(FATAL_ERROR "FATAL: In-source builds are not allowed. You should create a separate directory for build files.")
+endif()
+
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
```
jrmadsen, line 15:
```suggestion
list(INSERT CMAKE_MODULE_PATH 0 ${PROJECT_SOURCE_DIR}/cmake)
```

You should ensure that your `FindPAPI.cmake` is always picked up first.

CMakeLists.txt
jrmadsen: That's not very clear, especially with the `set(CMAKE_CXX_COMPILER ${Kokkos_CXX_COMPILER})` down below.

DavidPoliakoff: Yeah, could definitely use a comment. It's also more of a Lawrence Livermore-ism than something Sandia codes do today

CMakeLists.txt
MikolajZuzek: Fixed, thank you.

CMakeLists.txt
MikolajZuzek: Fixed, thank you (PAPI is now guarded with the option as suggested).

CMakeLists.txt
MikolajZuzek: Right, there's `USE_MPI` in source files. I'm working on that and will report back here.

CMakeLists.txt
MikolajZuzek: Great, thanks! Options prefixed.

cmake/FindPAPI.cmake
MikolajZuzek: Suggestion applied in b20828a09913424257c98cf33491de3199feb4f2

cmake/FindPAPI.cmake
MikolajZuzek: Suggestion applied in b20828a09913424257c98cf33491de3199feb4f2

cmake/FindPAPI.cmake
MikolajZuzek: Suggestion applied in b20828a09913424257c98cf33491de3199feb4f2

cmake/FindPAPI.cmake
MikolajZuzek: Suggestion applied, thank you.

I'll leave `if(TARGET PAPI::PAPI)` guard there, because we'd get target collisions in Apex or Caliper otherwise: they find PAPI using our find module now - since we prioritized it on `CMAKE_MODULE_PATH` (see the other applied remark) - so it can be called multiple times.

CMakeLists.txt
MikolajZuzek: I cleaned this up a bit (excuse messy draft). The idea is that:
1. we check if Kokkos is built within same parent project - and before Tools, as in David's example above. Is there a better way than checking for `TARGET Kokkos::kokkos` ? In this case we're all set up with options, compiler and flags.
2. we check for installed Kokkos in `Kokkos_ROOT` and extract all relevant settings from there.
3. in case Kokkos is nowhere found, we might just built Tools alone (figuring reasonable option defaults) or error out.

Any suggestions appreciated (where's the slicky part ?)

CMakeLists.txt
MikolajZuzek: Thanks. I started with the idea that we should strictly enforce same compiler and flags, then got course-corrected toward more flexible approach. I'm removing this for now, following your suggestion.

> Note: Instead of wiping or commenting that out I introduced `KokkosTools_REUSE_KOKKOS_COMPILER` advanced option as a way for the code to show our intentions (where the discussion stands).

CMakeLists.txt
MikolajZuzek: My motivation behind those macros was to move Caliper and Apex setup to the other file and keep scripts tight (applied in 4b7a3441904710e7c6e50952bc265eca83380b97) - let me know if you don't like that approach.

CMakeLists.txt
MikolajZuzek: (see macro discussion above)

CMakeLists.txt
MikolajZuzek: Added `KokkosTools_ENABLE_MPI` option in deb807bd501adc8626736d9f2e9ff700c10c6ad8 and based C++ `USE_MPI` on that (through config header).

profiling/chrome-tracing/CMakeLists.txt
MikolajZuzek: Thanks, applied this suggestion in deb807bd501adc8626736d9f2e9ff700c10c6ad8.
See MPI support discussion above.

CMakeLists.txt
MikolajZuzek: Thanks, fixed on 713ad680d32d52327e034c25201fe13fe0df5d52

CMakeLists.txt
MikolajZuzek: OpenMP is not actually used by tools, although it's mentioned in [`memory-usage` Makefile](https://github.com/kokkos/kokkos-tools/blob/7dc3f1996b0928fc8ac440c4d84cf62774dcbe35/profiling/memory-usage/Makefile#L2)
Cleaned that up on 713ad680d32d52327e034c25201fe13fe0df5d52

CMakeLists.txt
MikolajZuzek: I've changed all settings to cache variables and also minimized number of option defaults imposed on Caliper and Apex by the tools: it nails down to enabling support for Kokkos, MPI, Variorum, VTune, PAPI and CUDA/HIP support - when those are requested/detected (also the compiler if user enables `KokkosTools_REUSE_KOKKOS_COMPILER`).

MikolajZuzek:
Thank you @jrmadsen for a thorough review and great comments!

MikolajZuzek:
@crtrott @jrmadsen @daboehme @khuck @DavidPoliakoff @fnrizzi 

At this point I was able to draft a CMake build system for KokkosTools with:
- shared libraries built for most native tools and connectors;
- single library interface that connects all native tools, Caliper and some TPL connectors;
- Kokkos settings acquired from existing installation or Kokkos included in parent project;
- sample app that links tools and TPLs with new single library interface (see [`example`](https://github.com/NexGenAnalytics/kokkos-tools/tree/cmake-build-system/example) folder) + CMake tests running this demo with all implemented tools;
- Windows/MSVC build tested with VTune and Caliper;

You'll find more details in [the PR description](https://github.com/kokkos/kokkos-tools/pull/131#issue-1040253019) (just cleaned it up). I hope this shows overall idea about how things are connected, configured and built: any suggestions and ideas you may have will be much appreciated. My default plan is to continue with connecting the remaining tools same way.

crtrott:
FYI: I think this looks really good. I think priority should be the NVIDIA connector tool and the CMake export.

mzuzek:
@crtrott @dalg24 @fnrizzi 

Looks like we have the tasks listed in #128 covered to a decent extent.
Are there any immediate items you would see added or fixed on this PR ?

CMakeLists.txt
```diff
@@ -0,0 +1,220 @@
...
+
+message(STATUS)
+message(STATUS Configuring Kokkos-Tools)
+message(STATUS)
+
+# Common settings
+option(BUILD_SHARED_LIBS "Build shared libraries" ON)
+if(WIN32)
+  set(BUILD_SHARED_LIBS OFF) # We need to add __declspec(dllexport/dllimport) for Windows DLLs
+endif()
```
dalg24: What is the intent here?  `BUILD_SHARED_LIBS` is a global flag recognized by CMake.  It does not need to be an option.  Was that an attempt to change the default value to `ON`.
Also what is is the deal with Windows?  What would be the point to build static?

CMakeLists.txt
```diff
@@ -0,0 +1,220 @@
...
+
+# Tools settings
+option(KokkosTools_ENABLE_SINGLE  "Build single library interfacing all profilers and dispatching at runtime" OFF)
```
dalg24, line 30: Did we discuss that name "single"?  I suppose we can revisit later.

CMakeLists.txt
```diff
@@ -0,0 +1,220 @@
...
+option(KokkosTools_ENABLE_MPI     "Enable MPI support"              OFF)
+option(KokkosTools_ENABLE_CALIPER "Enable building Caliper library" OFF)
+option(KokkosTools_ENABLE_APEX    "Enable building Apex library"    OFF)
+option(KokkosTools_ENABLE_EXAMPLES "Build examples"                 OFF)
+# Advanced settings
+option(KokkosTools_REUSE_KOKKOS_COMPILER "Set the compiler and flags based on installed Kokkos settings" OFF)
+mark_as_advanced(KokkosTools_REUSE_KOKKOS_COMPILER)
+
+# Fetch Kokkos options:
+acquire_kokkos_config()
```
dalg24, line 45: When is this useful/desirable?

CMakeLists.txt
```diff
@@ -0,0 +1,220 @@
...
+  message(STATUS "PAPI support disabled")
+  set(KokkosTools_HAS_PAPI OFF)
+endif()
+
+if(KokkosTools_ENABLE_MPI)
+  find_package(MPI REQUIRED)
+  message(STATUS "Found MPI ${MPI_CXX_VERSION}: ${MPI_CXX_LIBRARIES}")
+  set(USE_MPI 1)
+else()
+  set(USE_MPI 0)
```
dalg24: OK for now but we will need to prefix these

CMakeLists.txt
```diff
@@ -0,0 +1,220 @@
...
+set(SINGLELIB_PROFILERS "" CACHE STRING "" FORCE)
+
+# Export settings
+include(GNUInstallDirs)
+set(EXPORT_NAME KokkosToolsConfig)
+set(EXPORT_INCLUDE_DIR ${CMAKE_INSTALL_INCLUDEDIR})
+set(EXPORT_LIB_DIR ${CMAKE_INSTALL_LIBDIR})
+set(EXPORT_TARGETS "" CACHE STRING "" FORCE)
+
+if(WIN32 OR OSX)
```
dalg24: OSX is not a CMake variable and the message below would be wrong anyway.

CMakeLists.txt
crtrott: Regarding the option: yeah I think the idea was to have the default be ON.
And on Windows you can still link a static tools library into your app, and use the API to set callbacks. We may wanna actually set the KokkosTools_ENABLE_SINGLE default to ON for windows.

CMakeLists.txt
crtrott: not sure we discussed but yeah I agree we can revisit. MONOLITHIC might be another option? Or do it the other way around and name the option KokkosTools_ENABLE_SEPARATE_LIBS or so, which by default is ON?

CMakeLists.txt
crtrott: Is this used for stuff like the connector tools to find the same CUDA library? Also there is some tools which have their own "ENABLE_CUDA" etc. e.g. the nvprof thing enabling is based on Kokkos_ENABLE_CUDA, and APEX use of CUPTI is based on the Kokkos option.

CMakeLists.txt
jrmadsen: If you want to get semi-lazy for windows and avoid all the import/export mess, just set [CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS](https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html) to ON.

CMakeLists.txt
vlkale: I think so, I will dig in.

CMakeLists.txt
vlkale: Using KokkosTools_ENABLE_MONOLITHIC makes most sense to me, JM2C.

CMakeLists.txt
vlkale: > Regarding the option: yeah I think the idea was to have the default be ON. And on Windows you can still link a static tools library into your app, and use the API to set callbacks. We may wanna actually set the KokkosTools_ENABLE_SINGLE default to ON for windows.

Yes, that makes sense. I will set KokkosTools_ENABLE_SINGLE (or KokkosTools_ENABLE_MONOLITHIC?) to ON.

CMakeLists.txt
vlkale: This is now fixed in NexGenAnalytics:cmake-build-system FYI.

CMakeLists.txt
vlkale: The `acquire_kokkos_config()` is used to show/print the kokkos configuration to the user, and to potentially set underlying (e.g.,CUPTI) tools options (e.g., ENABLE_CUDA).

vlkale: I can't find any connector tools right now making use of this from a quick look. I am leaving this line in for now.

CMakeLists.txt
vlkale: Removed.

vlkale: Removed in PR.

PR #132, opened 2021-11-23
Include data from regions & kernels that did not run on rank 0 in spa…

vbrunini:
…ce time stack.


crtrott:
Will this fall over if you run it with 10k ranks?

vbrunini:
I don't think it will for reasonable applications. We still only do the communication for any individual child name + kind pair a single time regardless of the number of ranks so unless an application does something strange like tack the rank onto the region/kernel name I think we'll be ok.

crtrott (approved):

PR #134, opened 2022-01-06
Disable global fencing for nvprof_connector tool.

vbrunini:
Since nvprof/nsys can appropriately time kernels without the fences
and it is desirable to be able to use them to investigate how well kernel
overlapping is working.

### DIFF
--- a/profiling/nvprof-connector/Makefile
+++ b/profiling/nvprof-connector/Makefile
@@ -9,7 +9,6 @@ all: kp_nvprof_connector.so
 MAKEFILE_PATH := $(subst Makefile,,$(abspath $(lastword $(MAKEFILE_LIST))))
 
 CXXFLAGS+=-I${MAKEFILE_PATH}
-CXXFLAGS+=-I${KOKKOS_ROOT}/include
 
 kp_nvprof_connector.so: ${MAKEFILE_PATH}kp_nvprof_connector.cpp
 	$(CXX) $(SHARED_CXXFLAGS) $(CXXFLAGS) $(LDFLAGS) \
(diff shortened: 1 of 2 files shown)


profiling/nvprof-connector/Makefile
```diff
@@ -9,6 +9,7 @@ all: kp_nvprof_connector.so
 MAKEFILE_PATH := $(subst Makefile,,$(abspath $(lastword $(MAKEFILE_LIST))))
 
 CXXFLAGS+=-I${MAKEFILE_PATH}
+CXXFLAGS+=-I${KOKKOS_ROOT}/include
```
dalg24, line 12: If I recall correctly we meant to avoid introducing a dependency on Kokkos.

profiling/nvprof-connector/kp_nvprof_connector.cpp
```diff
@@ -50,6 +50,11 @@
 #include <iostream>
 
 #include "nvToolsExt.h"
+#include "impl/Kokkos_Profiling_C_Interface.h"
```
dalg24, line 53: Should we define `Kokkos_Tools_ToolSettings` instead?

vbrunini: That seems very likely to lead to ODR violations from mis-matched kokkos & kokkos-tools versions at some point no?

dalg24: I would be surprised if it did actually give an issue unless Kokkos prepend another boolean option.

dalg24 (approved):

nliber (changes requested):
`padding` should be changed to `char` or `std::byte`, as `bool` is not guaranteed to be only one byte.  Also add a `static_assert` to make sure the size of the struct is always 256.

dalg24:
> `padding` should be changed to `char` or `std::byte`, as `bool` is not guaranteed to be only one byte. Also add a `static_assert` to make sure the size of the struct is always 256.

I agree but this would need to be updated on the Kokkos side first

nliber (approved):

PR #137, opened 2022-04-07
Create kp_sampler.cpp

vsurjadidjaja:
Proposing the addition of the LDMS-Kokkos sampler to the Kokkos-tools repo. This sampler is responsible for down regulating the amount of kernel function names sent to the LDMS-Kokkos connector. LDMS (Lightweight Distributed Monitoring System, https://github.com/ovis-hpc/ovis) provides a low overhead monitoring, transport and storage capability. LDMS uses Kokkos function calls to monitor application activity in real time. This sampler down regulates the number of kernel function calls from the application and send them over to LDMS's Kokkos connector. Makefile also available.


profiling/ldms-kokkos-sampler/kp_sampler.cpp
```diff
@@ -0,0 +1,206 @@
...
+       	}
+}
+
+extern "C" void kokkosp_push_profile_region(char* regionName) {
+
+}
+
+extern "C" void kokkosp_pop_profile_region() {
+
+}
```
dalg24: Why define them if they don't do anything?

profiling/ldms-kokkos-sampler/kp_sampler.cpp
```diff
@@ -0,0 +1,206 @@
...
+#include <inttypes.h>
+#include <execinfo.h>
+#include <cstdlib>
+#include <cstring>
+#include <map>
+#include <vector>
+#include <algorithm>
+#include <string>
+#include <iostream>
+#include <dlfcn.h>
```
dalg24: Include only what you need, so probably just
```suggestion
#include <cstdio>
#include <cstdint>
#include <cstdlib>
#include <cstring>
#include <dlfcn.h>
```

dalg24 (changes requested):
Include the Makefile from #138

profiling/ldms-kokkos-sampler/kp_sampler.cpp
vsurjadidjaja: Those specific functions are currently in discussion regarding how best to handle them. I can delete them since they currently aren't doing anything.

vsurjadidjaja:
Makefile has been added to this pull request and I have taken out the unnecessary lines of code.

nmhamster:
@vsurjadidjaja - the Kokkos sampler is not LDMS specific code so I would like to see if in a directory called `kokkos-sampler`. The benefit of this is that multiple tools can use the same sampler code. Can you make this change please?

profiling/ldms-kokkos-sampler/Makefile
```diff
@@ -0,0 +1,32 @@
+
+
+OVIS_DIR=/projects/ovis/streams/opt/ovis
```
nmhamster: The Kokkos sampler is not OVIS or LDMS specific so we should remove `OVIS` and `LDMS` references from the Makefile

profiling/ldms-kokkos-sampler/Makefile
```diff
@@ -0,0 +1,32 @@
...
+LDFLAGS=-L$(OVIS_DIR)/lib
+LIBS=-lldmsd_stream -lldms -lrt
+
+all: kp_kernel_ldms.so
+
+MAKEFILE_PATH := $(subst Makefile,,$(abspath $(lastword $(MAKEFILE_LIST))))
+
+CXXFLAGS+=-I${MAKEFILE_PATH}
+
+kp_kernel_ldms.so: ${MAKEFILE_PATH}kp_kernel_ldms.cpp ${MAKEFILE_PATH}kp_kernel_info.h
```
nmhamster: We only have the `kp_sampler.so` code in this PR so we need to remove references to the LDMS tools.

profiling/ldms-kokkos-sampler/Makefile
```diff
@@ -0,0 +1,32 @@
...
+	$(CXX) $(SHARED_CXXFLAGS) $(CXXFLAGS) $(LDFLAGS) -o $@ ${MAKEFILE_PATH}kp_kernel_ldms.cpp \
+		$(LIBS)	
+
+kp_sampler.so: ${MAKEFILE_PATH}kp_sampler.cpp ${MAKEFILE_PATH}kp_kernel_info.h
+	$(CXX) $(SHARED_CXXFLAGS) $(CXXFLAGS) $(LDFLAGS) -o $@ ${MAKEFILE_PATH}kp_sampler.cpp \
+		$(LIBS)	
+
+kp_sampler_sdh_vs.so: ${MAKEFILE_PATH}kp_sampler_sdh_vs.cpp ${MAKEFILE_PATH}kp_kernel_info.h
```
nmhamster: This was a previous debugging example so we should remove references to this sampler code.

nmhamster (commented):
@vsurjadidjaja - we need to remove the LDMS and OVIS references here and build the code on a standard Linux workstation to test the Makefile. I believe this would fail today because of the OVIS/LDMS references here. Can you remove these please?

dalg24 (changes requested):
Please move `profiling/ldms-kokkos-sampler/ -> common/kokkos-sampler/` so it is next to `kernel-filter`.  Eventually, it would be nice if these shared code instead of duplicating (not asking you to do this in this PR)
Remove the LDMS and OVIS mentions

dalg24 (commented):
It just occurred to me that this does not take into account deep copies, nor fences
https://github.com/kokkos/kokkos/blob/b6f691cb66c57071a27c5addb96cf7b8eb94fbd4/core/src/impl/Kokkos_Profiling_C_Interface.h#L112-L128
Was it intentional?

vsurjadidjaja:
All requested changes have been made.

common/kokkos-sampler/Makefile
```diff
@@ -0,0 +1,19 @@
+CXX=g++
+CXXFLAGS=-O3 -std=c++11 -g \
+	-I$(OVIS_DIR)/include/ -I./include
```
dalg24:
```suggestion
CXXFLAGS=-O3 -std=c++11 -g
```

common/kokkos-sampler/Makefile
```diff
@@ -0,0 +1,19 @@
...
+SHARED_CXXFLAGS=-shared -fPIC
+LDFLAGS=-L$(OVIS_DIR)/lib
+LIBS=-lldmsd_stream -lldms -lrt
```
dalg24: Reference to `ldmsd_stream` and `ldms` libs here

common/kokkos-sampler/Makefile
vsurjadidjaja: Thank you for pointing that out. They've been deleted now.

dalg24:
I had to do
```diff
diff --git a/common/kokkos-sampler/Makefile b/common/kokkos-sampler/Makefile
index 925103d..1414518 100644
--- a/common/kokkos-sampler/Makefile
+++ b/common/kokkos-sampler/Makefile
@@ -2,15 +2,13 @@ CXX=g++
 CXXFLAGS=-O3 -std=c++11 -g

 SHARED_CXXFLAGS=-shared -fPIC
-LIBS=-lrt

 MAKEFILE_PATH := $(subst Makefile,,$(abspath $(lastword $(MAKEFILE_LIST))))

 CXXFLAGS+=-I${MAKEFILE_PATH}

-kp_sampler.so: ${MAKEFILE_PATH}kp_sampler.cpp ${MAKEFILE_PATH}kp_kernel_info.h
-       $(CXX) $(SHARED_CXXFLAGS) $(CXXFLAGS) -o $@ ${MAKEFILE_PATH}kp_sampler.cpp \
-               $(LIBS)
+kp_sampler.so: ${MAKEFILE_PATH}kp_sampler.cpp
+       $(CXX) $(SHARED_CXXFLAGS) $(CXXFLAGS) -o $@ ${MAKEFILE_PATH}kp_sampler.cpp

 clean:
        rm *.so
```
to build and 
```diff
diff --git a/common/kokkos-sampler/kp_sampler.cpp b/common/kokkos-sampler/kp_sampler.cpp
index 8931cf0..6697910 100644
--- a/common/kokkos-sampler/kp_sampler.cpp
+++ b/common/kokkos-sampler/kp_sampler.cpp
@@ -104,7 +104,7 @@ extern "C" void kokkosp_init_library(const int loadSeq,
        }

        if( tool_verbosity > 0 ) {
-               printf("KokkosP: Sampling rate set to: %llu\n", tool_sample_rate );
+               printf("KokkosP: Sampling rate set to: %s\n", tool_sample_rate );
        }
 }
```
to avoid warnings
```
kokkos-tools/common/kokkos-sampler/kp_sampler.cpp:107:51: warning: format specifies type 'unsigned long long' but the argument has type 'const char *' [-Wformat]
                printf("KokkosP: Sampling rate set to: %llu\n", tool_sample_rate );
                                                       ~~~~     ^~~~~~~~~~~~~~~~
                                                       %s

**Comment by vsurjadidjaja** [comment]
Fixed.

File: common/kokkos-sampler/kp_sampler.cpp [hunk 8]
```diff
@@ -0,0 +1,192 @@
...
+
+	kernelCounter = 0;
+	uniqID = 1;
+
+	const char* tool_sample_rate = getenv("KOKKOS_SAMPLER_RATE");
+	if( NULL != tool_sample_rate ) {
+		kernelSampleRate = atoi( tool_sample_rate );
+	} else {
+		kernelSampleRate = 101;
+	}
```
dalg24, line 104: Actually now that I built the tool and gave it a spin, I find this counterintuitive. If I don't look at the implementation, when I select a "sampling rate" 2x bigger, I would expect to record twice more data but this does the opposite. 
Currently, `KOKKOS_SAMPLER_RATE=1` means record everything, `KOKKOS_SAMPLER_RATE=10` record every 10 kernels.

crtrott: Yeah its not a rate I guess.

crtrott: maybe rename to KOKKOS_SAMPLER_EVERY=10? As in "every 10th"? Or you could read in a fraction.

vsurjadidjaja:
Spoke to @crtrott and was advised to make a new file with changes to the variables that contain the word "rate" and replace it with "skip."

vsurjadidjaja:
Closing this PR to preserve the code under this branch. New PR submitted as PR#140. All change mentioned here have been carried over there.

vlkale:
> It just occurred to me that this does not take into account deep copies, nor fences https://github.com/kokkos/kokkos/blob/b6f691cb66c57071a27c5addb96cf7b8eb94fbd4/core/src/impl/Kokkos_Profiling_C_Interface.h#L112-L128 Was it intentional?

@vsurjadidjaja Following up from Damien's comment here, could you explain more on why this doesn't need to consider deep copies and fences? I think at the least deep copy should be considered, right?

PR #138, opened 2022-04-07
Makefile for kp_samper.cpp

vsurjadidjaja:
Adding the Makefile that is associated with appropriately compiling the kp_sampler.cpp file for the LDMS-Kokkos sampler.


vsurjadidjaja:
Merging this file into #137

PR #139, opened 2022-04-18
rename nvprof-connector to nvtx-connector, add 3 more hooks

cwpearson:
Changes / enhancements to nvprof-connector

1. rename nvprof-connector to nvtx-connector. nvprof is on its way out, and really this connector is to NVTX anyway.
2. Add hooks for `kokkosp_begin_fence`, `kokkosp_end_fence`, and `kokkosp_profile_event`


crtrott:
Can you add the code to disable auto fencing by the tool. I think we don't actually want it here. Furthermore, the fence profiling hooks are actually also getting called for the auto-fences, so any tool which profiles fencing needs to deal with that or get tons of false stuff.

cwpearson:
@crtrott is this the way to disable fencing?

https://github.com/cwpearson/kokkos-tools/blob/ad8848843e6fed0c0c3d76a8689b648eba44f3aa/profiling/nvtx-connector/kp_nvtx_connector.cpp#L60-L62

If so, I think we're good :D

vlkale:
> @crtrott is this the way to disable fencing?
> 
> 
> 
> https://github.com/cwpearson/kokkos-tools/blob/ad8848843e6fed0c0c3d76a8689b648eba44f3aa/profiling/nvtx-connector/kp_nvtx_connector.cpp#L60-L62
> 
> 
> 
> If so, I think we're good :D

@cwpearson yes, I believe that is right from looking at this a few months ago.

vlkale (commented):
I would suggest using 'anonymous' rather than the 'anon.' Abbreviation.

vlkale (approved):
LGTM. Thanks for these changes. Sorry it was somehow not reviewed until now. 

The fix to #147 (now in develop) may relate to or overlap with the discussion on the tools fence callback above.


I will need one other review before we can merge, and I have requested a reviewer.

vlkale (approved):

dalg24 (changes requested):
No PR against master.  You must target the develop branch.

vlkale:
@dalg24 Thanks. I will work with @cwpearson to fix.

cwpearson:
@dalg24 I have retargeted to  develop

vlkale:
> @dalg24 I have retargeted to develop

@cwpearson Thanks!

cwpearson:
Interesting, my local `clang-format-8` is happy with `profiling/nvtx-connector/kp_nvtx_connector.cpp` I'm not sure how to resolve the format-check failure.

vlkale:
> Interesting, my local `clang-format-8` is happy with `profiling/nvtx-connector/kp_nvtx_connector.cpp` I'm not sure how to resolve the format-check failure.

Strange. So, the `clang-format-8` was added in the last few months, and it should work through the CI in the same way that clang-format-8 works from the LLVM distribution containing clang-format-8. Can you provide the exact version info of clang-format-8? (The Kokkos team  has discussed changing the clang-format version in the Kokkos Tools CI from 8 to a higher one like 11, but that's a separate/orthogonal issue). 

@dalg24 @crtrott Do you know if there is an easy resolution to the failing format check here given results of CI?

cwpearson:
```bash
$ git log -s | head -n5
commit b3579f18cec82fe7b5dafcfff754823952e61457
Author: Carl Pearson <cwpears@sandia.gov>
Date:   Mon Apr 18 09:49:13 2022 -0700

    rename nvprof-connector to nvtx-connector, add hooks for kokkosp_begin_fence, kokkosp_end_fence, and kokkosp_profile_event
$ git status
On branch nvtx
Your branch is up to date with 'origin/nvtx'.

nothing to commit, working tree clean
$ clang-format-8 --version
clang-format version 8.0.1 (tags/RELEASE_801/final)
$ clang-format-8 -i profiling/nvtx-connector/kp_nvtx_connector.cpp
$ git status
On branch nvtx
Your branch is up to date with 'origin/nvtx'.

nothing to commit, working tree clean
```

cwpearson:
It would be great if the format check printed what it thought the diff should be! core and kernels do this

vlkale:
Thanks. I am a bit baffled here also. I am using version 8.0.1 and this works locally for me too (MacOS). Yes, that would be good if it did print what it should be, and I agree some fixes to format checking process needed for Kokkos Tools. I will circle back soon on this with either a fix to CI process involving clang-format-x or a (temporary) fix to that part of your code to get the format check passed.

profiling/nvtx-connector/kp_nvtx_connector.cpp
```diff
@@ -124,6 +124,21 @@ void kokkosp_stop_profile_section(const uint32_t sID) {
   nvtxRangeEnd(section.id);
 }
 
+void kokkosp_profile_event(const char* name) { nvtxMarkA(name); }
+
+void kokkosp_begin_fence(const char* name, const uint32_t deviceId,
+                         uint64_t* handle) {
+  if (nullptr == name) {
+    name = "anonymous. Kokkos fence";
+  }
```
dalg24: As far as I know this will never occur.

dalg24 (dismissed):
When profiling fences, we probably don't want kokkos to fence each time time it calls a callbacks.
https://github.com/kokkos/kokkos-tools/blob/0026352b197895a3d9a7c3f39fe65d8b315c3141/profiling/roctx-connector/kp_roctx_connector.cpp#L37-L40

We probably need a lever to toggle whether or not to enable fence profiling because this changes the behavior of the tool, either marking the full kernel execution or just the launch latency.

vlkale:
From trying this for another PR, I think this is fixed by including the .clang-format file at the top level of your PR directory (where, e.g., .gitignore, is). I don’t think I see that here?  (This is not obvious and all of that really should be documented in Contributing.md). Can you copy the one from develop into here?

profiling/nvtx-connector/kp_nvtx_connector.cpp
```diff
@@ -40,6 +40,24 @@ void kokkosp_init_library(const int loadSeq, const uint64_t interfaceVer,
   printf("KokkosP: NVTX Analyzer Connector (sequence is %d, version: %llu)\n",
          loadSeq, interfaceVer);
   printf("-----------------------------------------------------------\n");
+ const char* tool_global_fences = getenv("KOKKOS_TOOLS_NVTXCONNECTOR_GLOBALFENCES");
```
crtrott: We may actually just want to use KOKKOS_TOOLS_GLOBALFENCES?

vlkale: Changed to KOKKOS_TOOLS_GLOBALFENCES.

profiling/nvtx-connector/kp_nvtx_connector.cpp
```diff
@@ -0,0 +1,190 @@
...
+#include <cstdio>
+#include <cstdint>
+#include <vector>
+#include <string>
+
+#include "nvToolsExt.h"
+
+#include "kp_core.hpp"
+
+static int tool_globfences;  // use an integer
```
crtrott: what is this used for? Was maybe the intend to use that in the Kokkos_request_tool_settings which currently is a no-op? And if so should the default be true?   If that is the case you can init it here to 1, and then you don't need a else clause in the init.

crtrott (changes requested):

profiling/nvtx-connector/kp_nvtx_connector.cpp
```diff
@@ -0,0 +1,199 @@
...
+EXPOSE_PUSH_REGION(impl::kokkosp_push_profile_region)
+EXPOSE_POP_REGION(impl::kokkosp_pop_profile_region)
+EXPOSE_BEGIN_PARALLEL_FOR(impl::kokkosp_begin_parallel_for)
+EXPOSE_END_PARALLEL_FOR(impl::kokkosp_end_parallel_for)
+EXPOSE_BEGIN_PARALLEL_SCAN(impl::kokkosp_begin_parallel_scan)
+EXPOSE_END_PARALLEL_SCAN(impl::kokkosp_end_parallel_scan)
+EXPOSE_BEGIN_PARALLEL_REDUCE(impl::kokkosp_begin_parallel_reduce)
+EXPOSE_END_PARALLEL_REDUCE(impl::kokkosp_end_parallel_reduce)
+// TODO: expose section stuff
+}  // extern "C"
```
romintomasetti:
```suggestion
EXPOSE_END_PARALLEL_REDUCE(impl::kokkosp_end_parallel_reduce)
EXPOSE_CREATE_PROFILE_SECTION(impl::kokkosp_create_profile_section)
EXPOSE_START_PROFILE_SECTION(impl::kokkosp_start_profile_section)
EXPOSE_STOP_PROFILE_SECTION(impl::kokkosp_stop_profile_section)
}  // extern "C"
```

profiling/nvtx-connector/kp_nvtx_connector.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+  void kokkosp_end_fence(uint64_t handle) {
+    nvtxRangeId_t id = handle;
+    if (handle != std::numeric_limits<uint64_t>::max()) {
+      nvtxRangeEnd(id);
+    }
+  }
+
+}  // namespace NVProfConnector
+}  // namespace NVProfConnector
```
romintomasetti:
```suggestion
}  // namespace KokkosTools
```

profiling/nvtx-connector/kp_nvtx_connector.cpp
romintomasetti: I guess the goal is to allow the user to say "This profiling library requires fencing", using the environment variable `KOKKOS_TOOLS_GLOBALFENCES`. However, it seems other connectors don't have `KOKKOS_TOOLS_GLOBALFENCES`.

profiling/nvtx-connector/kp_nvtx_connector.cpp
```diff
@@ -0,0 +1,199 @@
...
+  char* profileLibrary = getenv("KOKKOS_TOOLS_LIBS");
+  if (NULL == profileLibrary) {
+    printf(
+        "Checking KOKKOS_PROFILE_LIBRARY. WARNING: This is a deprecated "
+        "variable. Please use KOKKOS_TOOLS_LIBS.\n");
+    profileLibrary = getenv("KOKKOS_PROFILE_LIBRARY");
+    if (NULL == profileLibrary) {
+      printf("KokkosP: No library to call in %s\n", profileLibrary);
+      exit(-1);
+    }
```
romintomasetti: I'm not sure this code is needed...

profiling/nvtx-connector/kp_nvtx_connector.cpp
```diff
@@ -0,0 +1,199 @@
...
+    my_event_set.push_region           = kokkosp_push_profile_region;
+    my_event_set.pop_region            = kokkosp_pop_profile_region;
+    my_event_set.begin_parallel_for    = kokkosp_begin_parallel_for;
+    my_event_set.begin_parallel_reduce = kokkosp_begin_parallel_reduce;
+    my_event_set.begin_parallel_scan   = kokkosp_begin_parallel_scan;
+    my_event_set.end_parallel_for      = kokkosp_end_parallel_for;
+    my_event_set.end_parallel_reduce   = kokkosp_end_parallel_reduce;
+    my_event_set.end_parallel_scan     = kokkosp_end_parallel_scan;
+    return my_event_set;
+  }
```
romintomasetti:
```suggestion
    my_event_set.end_parallel_scan     = kokkosp_end_parallel_scan;
    my_event_set.create_profile_section = kokkosp_create_profile_section;
    my_event_set.start_profile_section = kokkosp_start_profile_section;
    my_event_set.stop_profile_section = kokkosp_stop_profile_section;
    return my_event_set;
  }
```

romintomasetti (changes requested):
Hi! I went through the code, I think some slight and easy changes are required to go forward ;)

vlkale:
***@***.*** *Thanks for these. I will fix, test and get a reviewer for
merging.

profiling/nvtx-connector/kp_nvtx_connector.cpp
crtrott: Looks like this is now fixed good.

profiling/nvtx-connector/kp_nvtx_connector.cpp
crtrott: Yeah I don't think for now that this should do this, in particular you don't do anything with it anyway.

profiling/nvtx-connector/kp_nvtx_connector.cpp
crtrott: Yup this should be merged

profiling/nvtx-connector/kp_nvtx_connector.cpp
crtrott: yup that should be merged.

profiling/nvtx-connector/kp_nvtx_connector.cpp
crtrott: and yes those are missing

crtrott (changes requested):
Merge the suggested changes and remove the unnecessary code.

vlkale (approved):

vlkale:
This PR needs to resolve the conflict of the old file of kp_nvprof_connector.cpp on develop. Specifically, the nvtx branch on cwpearson/kokkos-tools needs to be synch'd with develop.

crtrott (approved):

crtrott:
@cwpearson can you review this

profiling/nvtx-connector/kp_nvtx_connector.cpp
```diff
@@ -23,12 +23,20 @@
...
+static int tool_globfences;  // use an integer
+
 namespace KokkosTools {
-namespace NVProfConnector {
+namespace NVTXConnector {
 
 void kokkosp_request_tool_settings(const uint32_t,
                                    Kokkos_Tools_ToolSettings* settings) {
-  settings->requires_global_fencing = false;
+  settings->requires_global_fencing = true;
```
crtrott, line 34: without this kernel renaming doesn't work, and the regions all only measure kernel launch

profiling/nvtx-connector/kp_nvtx_connector.cpp
dalg24: Why is it an `int` and not a `bool`?

profiling/nvtx-connector/kp_nvtx_connector_domain.h
dalg24, line 1: Why do we have a `_domain` suffix in the file name?

dalg24: Actually it even looks like it is never included so what is the point of that header?

romintomasetti:
Hi @dalg24 @crtrott !

Reading this PR, it urges us to add proper testing for Cuda, don't you think? :wink: (I mean, the `nvtx-connector` is compiled, but there is no guarantee that it actually works as intended :crab:)

profiling/nvtx-connector/kp_nvtx_connector.cpp
vlkale: > Why is it an `int` and not a `bool`?

My original reasoning was to provide extensibility to allow more modes of fencing. Talking with @crtrott, this is being changed to bool.

profiling/nvtx-connector/kp_nvtx_connector.cpp
```diff
@@ -18,13 +18,13 @@
...
-
+#include <limits>
 #include "nvToolsExt.h"
 
 #include "kp_core.hpp"
 
-static int tool_globfences;  // use an integer
-
+static int tool_globfences;  // use an integer for other options
+using namespace std;
```
dalg24: Why?  That's generally a bad idea.

PR #140, opened 2022-05-10
Addition of Kokkos sampler

vsurjadidjaja:
Proposing the addition of the LDMS-Kokkos sampler to the Kokkos-tools repo. This sampler is responsible for down regulating the amount of kernel function names sent to the LDMS-Kokkos connector. LDMS (Lightweight Distributed Monitoring System, https://github.com/ovis-hpc/ovis) provides a low overhead monitoring, transport and storage capability. LDMS uses Kokkos function calls to monitor application activity in real time. This sampler down regulates the number of kernel function calls from the application and send them over to LDMS's Kokkos connector. Makefile also available.

This is a resubmit of PR #137.


crtrott:
uhm isn't this all in the wrong places? i..e the sampler inside the filter, another Makefile fin some ldms-sampler directory?

vsurjadidjaja:
@crtrott Apologies for placing this in a sub directory. I have fixed the issue.

common/kokkos-sampler/kp_sampler_skip.cpp
```diff
@@ -0,0 +1,189 @@
...
+	void* deviceInfo) {
+
+	const char* tool_verbose_str = getenv("KOKKOS_SAMPLER_VERBOSE");
+	if( NULL != tool_verbose_str ) {
+		tool_verbosity = atoi( tool_verbose_str );
+	} else {
+		tool_verbosity = 0;
+	}
+
+	char* profileLibrary = getenv("KOKKOS_PROFILE_LIBRARY");
```
crtrott: This needs to check for KOKKOS_TOOLS_LIBS or both. And I would suggest error checks on getting nullptr back in case this wasn't set. For example someone could in theory compile the tool into their executable and use the callback setter interface to leverage this. In which case no environment variable is set or needs to be set.

crtrott (changes requested):

common/kokkos-sampler/kp_sampler_skip.cpp
vlkale: Thanks. Your revised code - with The warning of Kokkos_PROFILE_Library being deprecated - seems to addresses the comment.

I think you also may want to emit a warning - or note to the user - for the case of a null pointer where the tool is built with the executable. You would need to do this done only if the behavior of the tool when it is directly linked with the application program's (which includes performance and output of the logging) is significantly different than the case when the user specifies Kokkos_TOOLS_LIB environment variable. This may be nitpicking, but I wanted to make sure to point this out if helpful.

This looks good to me as it is.

vlkale:
LGTM with the most recent changes.

vlkale (approved):
LGTM.

dalg24 (changes requested):
PR needs to target the develop branch

vlkale:
> PR needs to target the develop branch

@vsurjadidjaja 

Following up here on @dalg24, do you think you can create a new PR, i.e., separate PR that is a copy of this one, which is against `develop` branch of Kokkos Tools github repository? The Kokkos team recently -  in the last few months - changed our developer contribution process  for Kokkos Tools such that contributions are no longer made to the master branch (which was the user-facing branch) and we will only provide the user-facing Kokkos Tools a release. On a related note: I am working to get Kokkos Tools released soon. Having your changes in the latest release will be useful so that it can be used by more applications.

Thanks!

vlkale:
@vsurjadidjaja Note that the clang-format 8 checks have failed

vsurjadidjaja:
Copied to PR#169 and built against the develop branch.

PR #149, opened 2022-10-19
SpaceTimeStack: Initialize/Finalize MPI if necessary 

### LINKED ISSUE #40: Profiling requires MPI be initialized
This might be a corner case. I have some tests in Trilinos that are single node only. They do not need MPI and, therefore, don't call MPI_Init in main (in cmake we restrict to one MPI process). However, if Trilinos is configured/built with MPI enabled, then it seems the stacked timer profiler requires all executables to initialize mpi. This causes a bunch of unit test failures in phalanx. An example is shown below. It would be convenient for the profiler could check if mpi was initialized and perform the init if needed. That way I could leave the profiling flag on during unit testing. When I unset the KOKKOS_PROFILE_LIBRARY flag, the error goes away and all the tests run fine.

```
[rppawlo@gge Kokkos]$ mpirun -np 1 ./Phalanx_tKokkosNestedLambda.exe 
Kokkos::OpenMP::initialize WARNING: OMP_PROC_BIND environment variable not set
  In general, for best performance with OpenMP 4.0 or better set OMP_PROC_BIND=spread and OMP_PLACES=threads
  For best performance with OpenMP 3.1 set OMP_PROC_BIND=true
  For unit testing set OMP_PROC_BIND=false
*** The MPI_Comm_rank() function was called before MPI_INIT was invoked.
*** This is disallowed by the MPI standard.
*** Your MPI job will now abort.
[gge.srn.sandia.gov:23656] Local abort before MPI_INIT completed completed successfully, but am not able to aggregate error messages, and not able to guarantee that all other processes were killed!
-------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code.. Per user-direction, the job has been aborted.
-------------------------------------------------------
--------------------------------------------------------------------------
mpirun detected that one or more processes exited with non-zero status, thus causing
the job to be terminated. The first process to do so was:

  Process name: [[60414,1],0]
  Exit code:    1
--------------------------------------------------------------------------
```

masterleinad:
Fixes https://github.com/kokkos/kokkos-tools/issues/40.
By default, `SpaceTimeStack` sets `USE_MPI=1` and assumes that the application uses MPI and initializes it. As also described in the link issue, this assumption is oftentimes not true. I mostly use the tool without MPI support anyway and always have to remember to use the correct flags to disable MPI support when building the tool. Just initializing `MPI` if it wasn't, seems like a good solution to both allow to run the tool in an MPI code or a non-MPI code without rebuilding.

### DIFF
--- a/profiling/space-time-stack/kp_space_time_stack.cpp
+++ b/profiling/space-time-stack/kp_space_time_stack.cpp
@@ -107,38 +107,44 @@ enum StackKind {
   STACK_COPY
 };
 
-void print_process_hwm() {
+void print_process_hwm(bool mpi_usable) {
   struct rusage sys_resources;
   getrusage(RUSAGE_SELF, &sys_resources);
   long hwm     = sys_resources.ru_maxrss;
   long hwm_max = hwm;
 
 #if USE_MPI
-  int rank, world_size;
-  MPI_Comm_rank(MPI_COMM_WORLD, &rank);
-  MPI_Comm_size(MPI_COMM_WORLD, &world_size);
+  if (mpi_usable) {
+    int rank, world_size;
+    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+    MPI_Comm_size(MPI_COMM_WORLD, &world_size);
 
-  // Max
-  MPI_Reduce(&hwm, &hwm_max, 1, MPI_LONG, MPI_MAX, 0, MPI_COMM_WORLD);
+    // Max
+    MPI_Reduce(&hwm, &hwm_max, 1, MPI_LONG, MPI_MAX, 0, MPI_COMM_WORLD);
 
-  // Min
-  long hwm_min;
-  MPI_Reduce(&hwm, &hwm_min, 1, MPI_LONG, MPI_MIN, 0, MPI_COMM_WORLD);
+    // Min
+    long hwm_min;
+    MPI_Reduce(&hwm, &hwm_min, 1, MPI_LONG, MPI_MIN, 0, MPI_COMM_WORLD);
 
-  // Average
-  MPI_Comm_size(MPI_COMM_WORLD, &world_size);
-  long hwm_ave;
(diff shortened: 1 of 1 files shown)


vlkale:
LGTM.

vlkale (commented):
LGTM.

dalg24:
I am not blocking it but this approach feels just wrong.  If MPI is not initialized then just do not attempt to use MPI routines.

jrmadsen:
> I am not blocking it but this approach feels just wrong. If MPI is not initialized then just do not attempt to use MPI routines.

FWIW, I agree, you should just surround all the MPI ops with the check for whether MPI is initialized. And maybe check whether MPI is initialized every time instead of storing the initial query during initialization -- I feel like there was somewhere in the docs that say you should initialize MPI before initializing kokkos but you could _potentially_ run into a situation where some other library used by the code initializes MPI after kokkos is initialized.

vlkale:
> I am not blocking it but this approach feels just wrong. If MPI is not initialized then just do not attempt to use MPI routines.

Is there a final resolution on this issue, given @dalg24 comment? This seems like a reasonable solution to me for now though I see @dalg24 point. Alternatively, we could just  put a warning/note kokkos documentation that one must be careful to not use MPI if it's initialized.

masterleinad:
> I am not blocking it but this approach feels just wrong. If MPI is not initialized then just do not attempt to use MPI routines.

The last commit addresses exactly that.

vlkale (approved):
LGTM. Thanks.

crtrott:
Doesn't this mean that you have to build against MPI now whereas we didn't before (with USE_MPI=0)?

masterleinad:
> Doesn't this mean that you have to build against MPI now whereas we didn't before (with USE_MPI=0)?

You can check for yourself that you can still add `-DUSE_MPI=0` as compile definition and build with a non-MPI compiler. All MPI calls are guarded with `USE_MPI`.

crtrott:
Ah my bad, I misperceived earlier the "ifdef USE_MPI" as being removed. Yeah this is fine then

PR #151, opened 2022-11-30
Fixing indents of Kokkos fences in Debugger 

vlkale:
Indents for debugger prints of each Kokkos::Fence call so that a user can easily identify/group the printed output for each fence more easily.


crtrott (changes requested):
I think I only wanted the names (in particular of parallel fors? extra indented? Isn't this a primary entry? Why should the primary fence entry be indented, I think its the names which need to be more indented so you can associate them.

PR #158, opened 2022-12-06
Format

crtrott:


.github/workflows/clang-format-check.yml
```diff
@@ -0,0 +1,13 @@
...
+jobs:
+  formatting-check:
+    name: Formatting Check
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Run clang-format style check for C/C++/Protobuf programs.
+      uses: jidicula/clang-format-action@v4.5.0
+      with:
+        clang-format-version: '13'
```
masterleinad: Why something else than what we are using in `kokkos/kokkos`?

crtrott: because our kokkos clang-format check runs on jenkins?

crtrott: I copied the one from desul over here.

masterleinad: My concern is rather that people have to use different `clang-format` versions to contribute to `Kokkos` repositories.

dalg24: that is an easy fix `s/13/14/`

masterleinad: Last I checked, we use `clang-format 8.0` in `kokkos/kokkos`. 🙂

dalg24: Oh right we changed on the ArborX side, not Kokkos...

vlkale: Yes, the clang-format-version of 8.0 in kokkos/kokkos should probably be upgraded to something like 13+. Otherwise, LGTM here.

vlkale (commented):
LGTM.

.github/workflows/clang-format-check.yml
masterleinad: I think it's less painful to change this to clang-format-8 since we are only starting to enforce indentation here.

vlkale (approved):

PR #159, opened 2022-12-16
Cmake build system 2

crtrott:

PR #163, opened 2023-02-27
Fix simple-kernel-timer Makefile

Rombur:
#159 introduced a `kp_shared.cpp` but we forgot to update the Makefile


dalg24 (approved):
Thanks for fixing

vlkale:
Thank you!

PR #169, opened 2023-03-21
Kokkos-LDMS

vsurjadidjaja:
Copy of PR#140. Built against develop branch as requested.


vlkale (approved):

vlkale:
@dalg24 I met with @vsurjadidjaja about the PR (now updated from #140 to #169). Regarding your comment on deep copy and fence from #140 and why it is isn't in the PR, I summarize my discussion with her. She told me that  this is so because getting data from these functions are not a critical requirement for the applications of relevance to their project, i.e., the Kokkos application program doesn't need LDMS sampling to be done. It will take time to get this additional functionality in there, and I understand it wouldn't be done in the next few weeks and would take longer than the time for a near-term release of Kokkos Tools. 

My reply to her was that a connector  should ideally support all tool callbacks to be generally useful for applications, and that we will definitely need to finish it later, but that not having it is OK given the interest and need for many Kokkos applications running on a variety of supercomputing platforms (particularly from the DoE) right now. I think it better to roll out a subset of tools callbacks (the majority of the important ones) in a Kokkos Tools connector that solidly work for Kokkos users than waiting until all of the tools callbacks are available. At least for this particular Kokkos Tools connector, the functionality for any of the other currently available tools callbacks, e.g., the kokkosp_begin/end_parallel_for does not depend on the tools callback of kokkosp_begin/end_deep_copy().  (This may not be the case for vendor tools like nvtx-connector).   

@vsurjadidjaja  correct me if I summarized anything wrong. 



@dalg24 cc. @crtrott What are your thoughts here? Are you OK with this PR being merged without the fence and deep_copy tools callbacks given the discussion summary above?

vlkale (approved):
I am good with this after discussion with @vsurjadidjaja. I have worked with her to put up a revised CMakeLists.txt file to incorporate the build of Kokkos Tools Connector (and I will be pushing that soon today or tomorrow).

vlkale:
Note that this is not passing clang-format-8 checks. The problem is in ./common/kokkos-sampler/kp_sampler_skip.cpp.

vlkale:
This needs an updated CMakeLists.txt file for building this connector. (I will be pushing an updated CMakeLists.txt file soon to @vsurjadidjaja's Kokkos-LDMS branch of her Kokkos Tools fork).

vlkale:
@crtrott cc. @dalg24  I fixed the clang-format check problem in ./common/kokkos-sampler/kp_sampler_skip.cpp by applying clang-format-8 on my laptop and the committing the resulting file.  I think this is ready to go.

CMakeLists.txt
```diff
@@ -45,6 +46,9 @@ option(KokkosTools_ENABLE_MPI     "Enable MPI support"              OFF)
 option(KokkosTools_ENABLE_CALIPER "Enable building Caliper library" OFF)
 option(KokkosTools_ENABLE_APEX    "Enable building Apex library"    OFF)
 option(KokkosTools_ENABLE_EXAMPLES "Build examples"                 OFF)
+option(KokkosTools_ENABLE_LDMSSAMPLER "Enable Kokkos LDMS Sampler" OFF)
```
dalg24: Why would the sampler gets its own option and not be always enabled?

cmake/configure_ldmssampler.cmake
dalg24, line 1: Why did you add this empty file?

dalg24 (commented):
What's LDMS and why is it a relevant part in the name of the tool being added?

vlkale:
@dalg24 LDMS stands for lightweight data monitoring system. LDMS is a Sandia-led project and tool for collecting logging and profiling data (as well as potentially some higher-level information) about the behavior of several different jobs, e.g., scientific simulation, HPC benchmark, simple script for a ping pong test, run on a particular supercomputer.

@vsurjadidjaja can the above be incorporated into the kokkos Tools sampler documentation? Let me know what you think and if I've described it properly.

Because LDMS works best with samples of profiling and not all profiling data of jobs run on a supercomputer, the Kokkos Tools sampler connector (which I think ought to be called kernel sampler) was built to provide capabilities of _sampling_  timings or profiling data of Kokkos kernels in an Kokkos application using a particular sampling rate provided by the user or system administrator. The sampler sits on top of another Kokkos tool, e.g., simple kernel timer. If sampling is turned on and configured with a sampling rate of 10%: when a kokkos application of 40 parallel_for kernels is run with, e.g., simple kernel timer, the user will see the timings for that parallel_for kernel at invocation number 9, invocation 19, invocation 29, invocation 39. These timings get fed to The Kokkos tools sampler is actually not intended to be specific to giving its samples to LDMS. However, at the moment it hasn’t been tested without it. Having thought about this and looked more carefully at the sampler code, it is decoupled from LDMS, I think the LDMS sampler Kokkos Tools option should be renamed to sampler. @vsurjadidjaja  and I will change this.

CMakeLists.txt
vlkale: In terms of getting its own option, there may be cases when a user doesn’t want sampling of data and wants data from every invocation. Then, sampling should be turned off. Whether or not we get rid of the option, I would like to propose we make the sampling rate a Kokkos Tools cmake configurable _value_ option. If we do provide sampling as an Kokkos Tools cmake option, the default option value for sampling rate can be, say, 50%. Note that 0% means no profiling or logging is done. If we don’t provide sampling as an Kokkos Tools cmake option, the default configurable value option is 100%.

cmake/configure_ldmssampler.cmake
vlkale: I thought it was needed for the sampler, and I was going to work with Vanessa to put options here for the sampler. It looks like this wasn't done. Anyway, I think it is now actually not needed, and I will delete, making sure once with @vsurjadidjaja that is good with her.

crtrott:
The other point of Damien was that we don't need an option for the sampler. It doesn't depend on anything. All the other options are not so much for the tool, but enabling the relevant TPL such as Apex and Caliper

vlkale:
> It doesn't depend on anything. All the other options are not so much for the tool, but enabling the relevant TPL such as Apex and Caliper 

Thanks for your clarifying note and I see we must get rid of the option.

vlkale:
> > It doesn't depend on anything. All the other options are not so much for the tool, but enabling the relevant TPL such as Apex and Caliper 
> 
> Thanks for your clarifying note and I see we must get rid of the option.  
> 

I think there is another related question that remains here, as has become apparent to me as I was writing in my reply to @dalg24 comment on this PR: Does the Kokkos application programmer configure/set the sampling rate to this tool when using this tool? Or, should it be automatic and/or be set by the supercomputer system administrator?

Here are two follow-on questions if the answer to the first question is yes (not mentioned in my reply to @dalg24 earlier), and this is also for Vanessa to look into if she can:

* Through what means does a user set the sampling rate? Should this be a Kokkos Tools environment variable and/or command-line argument to be parsed, in a similar fashion as that of Kokkos Tools PR #171?

* How does the user _know_ what sampling rate is best to use for her on application on a given platform to minimize profiling overhead and storage, for best behavior of the application and supercomputing system as a whole? Perhaps we just explain or advise what the sampling rate ought to be (considering different kinds of applications and different systems, and considering the needs of profiling and monitoring to be done) through a help message via the Kokkos Tools Kokkosp_help function?

cmake/configure_ldmssampler.cmake
vlkale: FYI, this is taken out now.

vlkale:
Went through environment variable addition with @vsurjadidjaja and it looks good to her.

common/kokkos-sampler/kp_sampler_skip.cpp
```diff
@@ -0,0 +1,239 @@
...
+static initFunction initProfileLibrary         = NULL;
+static finalizeFunction finalizeProfileLibrary = NULL;
+static beginFunction beginForCallee            = NULL;
+static beginFunction beginScanCallee           = NULL;
+static beginFunction beginReduceCallee         = NULL;
+static endFunction endForCallee                = NULL;
+static endFunction endScanCallee               = NULL;
+static endFunction endReduceCallee             = NULL;
+
+extern "C" void kokkosp_init_library(const int loadSeq,
```
crtrott: all these extern "C" here need to go. The expose macros do that.

crtrott (changes requested):

common/kokkos-sampler/kp_sampler_skip.cpp
```diff
@@ -0,0 +1,239 @@
+#include <cstdio>
+#include <cstdint>
+#include <cstdlib>
+#include <cstring>
+#include <dlfcn.h>
+
+namespace KokkosTools {
+namespace Sample {
```
crtrott: make that Sampler

common/kokkos-sampler/kp_sampler_skip.cpp
```diff
@@ -0,0 +1,239 @@
...
+      }
+    }
+  }
+
+  free(envBuffer);
+
+  kernelCounter = 0;
+  uniqID        = 1;
+
+  const char* tool_sample = getenv("KOKKOS_TOOLS_SAMPLE_SKIP");
```
crtrott: this should be `KOKKOS_TOOLS_SAMPLER_SKIP` so we are consistent.

crtrott (changes requested):

common/kokkos-sampler/kp_sampler_skip.cpp
vlkale: I have changed this, also updating the Kokkos Tools wiki name of this tool back to sampler and not sample.

common/kokkos-sampler/kp_sampler_skip.cpp
vlkale: Done.

vlkale:
Per discussion last week and this week with @crtrott @vsurjadidjaja  this is a _utility_ and not a tool by itself. So, it should be used in conjunction with  (linked with) a profiling or debugging tool in the subdirectory profiling or debugging of the Kokkos Tools repo, e.g., the simple kernel timer.

crtrott (approved):

dalg24:
Per #182 this does not build.
Fix or revert ASAP.

vlkale:
@dalg24 Working on this fix now. I will revert if it isn't fixed within the hour. 

@vsurjadidjaja

PR #171, opened 2023-03-28
Space Time Stack: Adding --kokkos-tools-args for the output threshold

### LINKED ISSUE #108: space-time-stack: configurable threshold for filtering out kernels
Currently we have an arbitrary threshold for what we don't print out (0.1%), which should be a configurable value

csiefer2:
Closes #108

### DIFF
--- a/profiling/space-time-stack/kp_space_time_stack.cpp
+++ b/profiling/space-time-stack/kp_space_time_stack.cpp
@@ -30,7 +30,6 @@
 #include <sys/resource.h>
 #include <algorithm>
 #include <cstring>
-#include <stdlib.h>
 
 #include "kp_core.hpp"
 
@@ -42,6 +41,9 @@
 
 namespace KokkosTools {
 namespace SpaceTimeStack {
+
+// Threshold to use for output (can be set via CLI options)
+double output_threshold = 0.1;
 
 enum Space { SPACE_HOST, SPACE_CUDA, SPACE_HIP, SPACE_SYCL, SPACE_OMPT };
 
@@ -245,11 +247,7 @@ struct StackNode {
     static bool add_comma = false;
     auto percent          = (total_runtime / tree_time) * 100.0;
 
-    double threshold        = 0.1;
-    const char* s_threshold = getenv("KOKKOSP_PRINT_THRESHOLD");
-    if (s_threshold) threshold = strtod(s_threshold, NULL);
-
-    if (percent < threshold) return;
+    if (percent < output_threshold) return;
     if (!name.empty()) {
       if (add_comma) os << ",\n";
       add_comma = true;
@@ -332,11 +330,7 @@ struct StackNode {
                        double tree_time) const {
     auto percent = (total_runtime / tree_time) * 100.0;
 
-    double threshold        = 0.1;
-    const char* s_threshold = getenv("KOKKOSP_PRINT_THRESHOLD");
-    if (s_threshold) threshold = strtod(s_threshold, NULL);
-
-    if (percent < threshold) return;
+    if (percent < output_threshold) return;
     if (!name.empty()) {
       os << my_indent;
       auto imbalance = (max_runtime / avg_runtime - 1.0) * 100.0;
@@ -833,6 +827,43 @@ Kokkos::Tools::Experimental::EventSet get_event_set() {
   my_event_set.end_parallel_scan     = kokkosp_end_parallel_scan;
   return my_event_set;
 }
+
+
+ extern "C" {
+   void kokkosp_print_help(const char* exe)
+   {
+     auto usage = R"usage(
+Default value: 0.1
+
+Description:
+  Provide a decimal threshold value of percent of parent time for output.  
+  Timers below this threshold will not be output.  Set to 0 to get unfiltered
+  reports.
+
+Example:
+  The following example would set the threshold to 10%
+    <exe> [--kokkos-tools-args 10 ]
+)usage";
+     std::cout << "usage: " << exe << "[--kokkos-tools-args <threshold>]\n" << usage;
+   }
+
+   void kokkosp_parse_args(int argc, char** argv)
+   {
+     // See description in original PR.
+     // argc will always be at least 1 (exe)
+     if(argc == 1) {
+       // No argument, use the default
+     }
+     else if(argc == 2) {
+       // User specified a threshold
+       output_threshold = strtod(argv[1],0);
+     }
+     else { 
+       // Too many args
+       kokkosp_print_help(argv[0]); exit(1); 
+     }
+   }
+ };//extern C
 
 }  // namespace SpaceTimeStack
 }  // namespace KokkosTools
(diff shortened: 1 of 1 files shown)


csiefer2:
@vlkale Got another one...

dalg24 (commented):
Please discuss specifying as a double VS an integer (in percent)

csiefer2:
> Please discuss specifying as a double VS an integer (in percent)

@dalg24 The code's default behavior is to include anything above 0.1%.  You can't get that behavior with an integer and having an option that can't reproduce the default behavior is a little weird.

vlkale:
> > Please discuss specifying as a double VS an integer (in percent)
> 
> 
> 
> @dalg24 The code's default behavior is to include anything above 0.1%.  You can't get that behavior with an integer and having an option that can't reproduce the default behavior is a little weird.
> 

Yes, I agree that is weird. I think a double could be OK, though perhaps we use 0-1000 to represent a 10th of a percent or something? On a separate note: Perhaps some documentation is needed on why the Kokkos Tools developers choose 0.1% as the default.

I guess maybe the bigger question here to consider - which may actually be behind @dalg24 comment - is: 

(1) how _much granularity_ of control this and possibly any Kokkos Tool should give to the user for the knob on the amount of profiling and debugging information emitted

(2) how this or any Kokkos Tool should _interpret_ the knob value set by the user, considering an application run on any Kokkos backend. Does .00001% mean anything for a very short program? If it does, is it at all possible it may mean different things for different backends, e.g., CUDA, HIP, OpenMPTarget?

csiefer2:
@vlkale Seeing as the threshold isn't documented in the wiki at all (https://github.com/kokkos/kokkos-tools/wiki/Space-Time-Stack) these are very good questions :)

vlkale:
Thanks. :) documenting this better is an action item for me.

jrmadsen:
Just FYI bc environment variables are generally so opaque/hidden, I added the ability for Kokkos tools to accept command line arguments a long time ago:

https://github.com/kokkos/kokkos/pull/3627

vlkale:
@jrmadsen Ah, thanks for reminding of that. We ought to keep that in mind for this PR.

csiefer2:
@jrmadsen 

So the Kokkos tool just needs to implement kokkosp_parse_args and then go and grab the relevant arguments?

jrmadsen:
> @jrmadsen
> 
> So the Kokkos tool just needs to implement kokkosp_parse_args and then go and grab the relevant arguments?

Yes, but you should also implement `kokkosp_print_help` to specify what the argument(s) mean.

Probably would look something like this:

```cpp
void kokkosp_print_help(const char* exe)
{
     auto usage = R"usage(
Default value: ...

Description:
  Provide a decimal threshold value for ...

Example:
  The following example would set the threshold to 10%
    <exe> --kokkos-tools-args 10 ...
)usage";
     std::cout << "usage: " << exe << " --kokkos-tools-args <threshold>\n" << usage;
}

void kokkosp_parse_args(int argc, char** argv)
{
    // See description in original PR.
    // argc will always be at least 1 (exe)
    if(argc == 2) threshold = strtod(argv[1]);
    else { kokkosp_print_help(argv[0]}; exit(1); }
}

**Comment by jrmadsen** [comment]
Documenting the environment variable in the print_help function would probably be a good way for these libs to "self-document" their env controls too

**Comment by csiefer2** [comment]
@vlkale I can add a CLI version of that control as well.  If I do that, would that be good enough for you?  At which point we can bug @dalg24 and @crtrott ?

**Comment by vlkale** [comment]
Yes, I think that will be great, and I think the code by @jrmadsen on the help text for the command-line arguments is very useful. I think that how the parameter should be represented could then be iterated on with @dalg24 and @crtrott.

**Comment by vlkale** [comment]
> Just FYI bc environment variables are generally so opaque/hidden, I added the ability for Kokkos tools to accept command line arguments a long time ago:
> 
> 
> 
> https://github.com/kokkos/kokkos/pull/3627

Following up on this: 

We can and ought to leave out using environment variable for this PR (or at least an initial version), as I think suggested through @jrmadsen’s feedback here. 

I want to add a few more thoughts of mine on this here: I think real-time/online adjustments to choices and configuration of the tool ought to be allowed, say for a scientific workflow application. I do think the having the corresponding environment variable approach is also useful. This environment can be can extracted within the callback functions of Kokkos tool connector using a function like getenv(). For this particular configuration choice of stopping threshold of aggregated output over the lifetime of the application, I can’t think of a situation at the moment where online selection doesn’t make sense. The point is that the approach of specifying runtime configuration values through a command-line flag shouldn’t mean forgoing specifying parameters through environment variables.

**Comment by vlkale** [comment]
> Thanks. :) documenting this better is an action item for me.

This is done, and it is in the SpaceTimeStack wiki.

**Comment by csiefer2** [comment]
@vlkale I keep getting this "Tool has not provided a help message"

Is there something I have to do to "register" the CLI option, e.g. like the EXPOSE macros that get used for other things?

**Comment by jrmadsen** [comment]
> @vlkale I keep getting this "Tool has not provided a help message"
> 
> Is there something I have to do to "register" the CLI option, e.g. like the EXPOSE macros that get used for other things?

Any chance you forgot the `extern "C"` that I left out of my snippet? That would certainly cause the dlsym lookup to fail.

**Comment by vlkale** [comment]
> > @vlkale I keep getting this "Tool has not provided a help message"
> 
> > 
> 
> > Is there something I have to do to "register" the CLI option, e.g. like the EXPOSE macros that get used for other things?
> 
> 
> 
> Any chance you forgot the `extern "C"` that I left out of my snippet? That would certainly cause the dlsym lookup to fail.

@csiefer2 I am looking this up with respect to the message you are getting. I think adding extern  "C" as @jrmadsen says should be done whatever the case. I am looking into the EXPOSE macro more and will get back to you soon.

**Comment by jrmadsen** [comment]
I'm not sure what the references to the EXPOSE macros are about but from these two snippets, it's seems pretty clear that the dlsym lookup for the `kokkosp_print_help` function is failing:

https://github.com/kokkos/kokkos/blob/aa1f48f3172069f212ed2c97b74b786999b26f17/core/src/impl/Kokkos_Profiling.cpp#L192

https://github.com/kokkos/kokkos/blob/aa1f48f3172069f212ed2c97b74b786999b26f17/core/src/impl/Kokkos_Profiling.cpp#L508

**Comment by csiefer2** [comment]
@jrmadsen @vlkale Fixes pushed.  Let me know what you think!

**Comment by e10harvey** [comment]
Consider replacing global state (environment settings) with a command line option for each kokkos tool: `--settings=/path/to/settings.{yaml,json}`.

**Comment by csiefer2** [comment]
> Consider replacing global state (environment settings) with a command line option for each kokkos tool: `--settings=/path/to/settings.{yaml,json}`.

That's basically what this PR does, for the one variable this tool uses.

**Review by crtrott — APPROVED** [review]
Looks good for now. We should discuss whether we want to add a KOKKOS_TOOLS_SPACETIMESTACK_THRESHOLD environment variable?

**Comment by jrmadsen** [comment]
> Looks good for now. We should discuss whether we want to add a KOKKOS_TOOLS_SPACETIMESTACK_THRESHOLD environment variable?

Maybe just initialize it based on the environment variable, if it exists, so the command line can override it (since the latter is far more explicit):

```cpp
double output_threshold = getenv("...") ? strtod(getenv("..."), nullptr) : 0.1;
```

PR #184, opened 2023-04-26
Fix roctx-connector build system

Rombur:
We can't compile the roctx connector with CMake (currently there are no CMakeLists.tx). There no equivalent to `Find_package(CUDAToolKit)` with ROCm, instead I use the $ROCM_PATH environment variable as a hint. I've tested this on Frontier


CMakeLists.txt
```diff
@@ -185,8 +185,8 @@ if(Kokkos_ENABLE_CUDA)
   add_subdirectory(profiling/nvprof-connector)
   add_subdirectory(profiling/nvprof-focused-connector)
 endif()
-if(KOKKOS_ENABLE_HIP)
-  #add_subdirectory(profiling/roctx-connector)
+if(Kokkos_ENABLE_HIP)
```
dalg24, line 188: Did you consider enabling if ROCm can be found?

dalg24 (approved):

CMakeLists.txt
Rombur: The problem with ROCm is that CMake wants you to enable it has a language. I don't have an easy check like CUDA where you can use `Find_package(CUDAToolKit)`. That's why I did it this way

crtrott (approved):

PR #185, opened 2023-04-26
Update kp_sampler_skip.cpp

vlkale:
Added Kp core header file   needed for interface with Kokkos (as Expose Tool settings not found). This is aimed to fix problem with building of PR #169.

### DIFF
--- a/common/kokkos-sampler/kp_sampler_skip.cpp
+++ b/common/kokkos-sampler/kp_sampler_skip.cpp
@@ -3,7 +3,7 @@
 #include <cstdlib>
 #include <cstring>
 #include <dlfcn.h>
-#include <kokkos_core.hpp>
+#include <kp_core.hpp>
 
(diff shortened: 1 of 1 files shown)


common/kokkos-sampler/kp_sampler_skip.cpp
```diff
@@ -3,6 +3,7 @@
 #include <cstdlib>
 #include <cstring>
 #include <dlfcn.h>
+#include <kokkos_core.hpp>
```
dalg24: That header file does not exist and even if you spelled it properly that is not the right fix.

vlkale (commented):
I think this should actually be #include <kp_core.hpp> having looked at the code files along with the PR comments. I am testing this.

vlkale:
Redirecting to PR #187. This PR #185 was created before the associated revert via #186.

PR #187, opened 2023-04-26
Fix and improve sampler

### LINKED ISSUE #180: Sampler needs to allow for randomized sampling (not just periodic)
# Problem 

Currently, the Kokkos Tools Sampler utility only allows for a periodic sampling and not both periodic and random sampling. 

# Why Important 

Limiting to periodic sampling is inconvenient for users to think about given a very small number of kernel invocations relative to the periodicity (skip 'rate' in the sampler tool). 

# Solution 

The Kokkos Tool sampler utility needs to allow for the user to choose randomized sampling and periodic sampling. 

# Notes 
Perhaps a subsequent topic is to allow the user to choose and/or configure some composition of both. This could be left for a later PR.

vlkale:
* Fix with kp_core.hpp and kp_config.hpp. This allows for enhancement for Kokkos Tools fences.


vlkale:
This builds now. One note is that I have been getting warnings on mac osx for snprintf (as before). I am leaving this as is, but squashing this warning may help here  (especially given discussions on printf's in Kokkos)?

vlkale:
[outputForBuild-kokkos-sampler.pdf](https://github.com/kokkos/kokkos-tools/files/11338974/outputForBuild-kokkos-sampler.pdf)

masterleinad:
> especially given discussions on printf's in Kokkos)?

We were talking about using `printf` in kernels.

vlkale:
Got it, thanks On Apr 26, 2023, at 10:29 PM, Daniel Arndt ***@***.***> wrote:﻿

especially given discussions on printf's in Kokkos)?

We were talking about using printf in kernels.

masterleinad:
I think we can fix warning in other pull requests.

common/kokkos-sampler/README.md
```diff
@@ -0,0 +1,4 @@
+This is a sampler utility that is intended to complement other tools in the Kokkos Tools set. This utility allows for sampling (rather than collecting) of profiling or debugging data gathered from a particular tool of the Kokkos Tools set. The Kokkos Tools user provides asampling rate via the environment variable KOKKOS_TOOLS_SAMPLER_SKIP.
```
masterleinad:
```suggestion
This is a sampler utility that is intended to complement other tools in the Kokkos Tools set. This utility allows for sampling (rather than collecting) of profiling or debugging data gathered from a particular tool of the Kokkos Tools set. The Kokkos Tools user provides a sampling rate via the environment variable KOKKOS_TOOLS_SAMPLER_SKIP.  
```

common/kokkos-sampler/README.md
```diff
@@ -0,0 +1,4 @@
+This is a sampler utility that is intended to complement other tools in the Kokkos Tools set. This utility allows for sampling (rather than collecting) of profiling or debugging data gathered from a particular tool of the Kokkos Tools set. The Kokkos Tools user provides asampling rate via the environment variable KOKKOS_TOOLS_SAMPLER_SKIP.  
+
+Tool fencing - which ensures state of sampled profiling and logging data in memory is captured at the time of the callback invocation - is important to control for performance of this utility while also ensuring accurate data. This utility handles this. 
+However, turning on tool fencing for this to work is necessary for this to work. Turn on this fencing by setting the environment variable `KOKKOS_TOOLS_GLOBALFENCES` to be `1`.
```
masterleinad: I find this very confusing to read. Maybe,
```suggestion
In order for the state of the sampled profiling and logging data in memory to be captured at the time of the utility's callback invocation, it might be important to enforce fences. However, this also means that there are more synchronization points compared with running the program without the tool.
This fencing behavior can be controlled by setting the environment variable `KOKKOS_TOOLS_GLOBALFENCES`. A non-zero value implies global fences on invocation of the tool. The default is not to introduce extra fences.	
```
?

vlkale: Thanks for those suggestions, and I can see where it is confusing. One question: why _might_ enforcing [global/tool-induced]fences be important at the point at which the tool callback is invoked? In what situation is not enforcing global/tool-induced fencing needed? (One is precisely on a callback invocation for which data gathering is skipped due to sampling). Do you think that the word 'might' introduces ambiguity? Can that word _might_ be _is_?

> In order for the state of the sampled profiling and logging data in memory to be captured at the time of the utility's callback invocation, it might be important to enforce fences.

masterleinad: > Thanks for those suggestions, and I can see where it is confusing. One question: why _might_ enforcing [global/tool-induced]fences be important at the point at which the tool callback is invoked? In what situation is not enforcing global/tool-induced fencing needed? (One is precisely on a callback invocation for which data gathering is skipped due to sampling). Do you think that the word 'might' introduces ambiguity? Can that word _might_ be _is_?

If we say "is", I would expect that (extra) fences are always used.

vlkale: OK, I think having the word _is_ this is fine as it is. 

For grammar, I would just change this 
>> it might be important to enforce fences.

to 

>  enforcing fences might be important.


If there are any other comments or clarifications on the tool utility purpose and functionality, let me know. Otherwise, LGTM.

crtrott (approved):

PR #188, opened 2023-04-27
ci: build kokkos tools with kokkos

romintomasetti:
As promised in #182, here is a better version of the CI, that builds against `OpenMP`, `Cuda` and `HIP` backends.

Things to discuss:
- [x] Do we keep the "simple CI" version? -> Yes, it is important to keep a "standalone" build.
- [x] Do you agree with the CMake presets based stuff?


kokkos.presets.json
```diff
@@ -0,0 +1,68 @@
...
+        "patch" : 0
+    },
+    "configurePresets" : [
+        {
+            "name"          : "default",
+            "displayName"   : "default",
+            "binaryDir"     : "${sourceDir}/build-with-${presetName}",
+            "cacheVariables" : {
+                "CMAKE_EXPORT_COMPILE_COMMANDS" : {"type" : "BOOL"  , "value" : "ON"},
+                "CMAKE_BUILD_TYPE"              : {"type" : "STRING", "value" : "RELEASE"},
```
dalg24: Isn't it spelled "Release"?

kokkos.presets.json
```diff
@@ -0,0 +1,68 @@
...
+        },
+        {
+            "name"        : "GNU-Cuda",
+            "displayName" : "GNU-Cuda",
+            "inherits"    : "default",
+            "cacheVariables" : {
+                "Kokkos_ENABLE_CUDA"        : {"type" : "BOOL"  , "value" : "ON"},
+                "Kokkos_ENABLE_CUDA_LAMBDA" : {"type" : "BOOL"  , "value" : "ON"},
+                "Kokkos_ARCH_VOLTA70"       : {"type" : "BOOL"  , "value" : "ON"},
+                "CMAKE_CXX_COMPILER"        : {"type" : "STRING", "value" : "$penv{PWD}/bin/nvcc_wrapper"}
```
dalg24: Why are you using `penv${PWD}` instead of just `env${PWD}`?

crtrott:
I think we should keep the "simple CI" since it tests a build without Kokkos Core present which I'd like to keep going for simple tools.

Otherwise the preset stuff looks ok to me, any other thoughts?

kokkos.presets.json
```diff
@@ -0,0 +1,68 @@
...
+    "version"   : 3,
+    "cmakeMinimumRequired" : {
+        "major" : 3,
+        "minor" : 22,
+        "patch" : 0
+    },
+    "configurePresets" : [
+        {
+            "name"          : "default",
+            "displayName"   : "default",
```
dalg24: What's the point of using `displayName` if we are just echoing the name of the preset?

kokkos.presets.json
```diff
@@ -0,0 +1,68 @@
...
+            "binaryDir"     : "${sourceDir}/build-with-${presetName}",
+            "cacheVariables" : {
+                "CMAKE_EXPORT_COMPILE_COMMANDS" : {"type" : "BOOL"  , "value" : "ON"},
+                "CMAKE_BUILD_TYPE"              : {"type" : "STRING", "value" : "RELEASE"},
+                "CMAKE_CXX_STANDARD"            : {"type" : "STRING", "value" : "17"},
+                "CMAKE_CXX_EXTENSIONS"          : {"type" : "BOOL"  , "value" : "OFF"},
+                "BUILD_SHARED_LIBS"             : {"type" : "BOOL"  , "value" : "ON"}
```
dalg24: Does it matter?

kokkos.presets.json
```diff
@@ -0,0 +1,68 @@
...
+            }
+        },
+        {
+            "name"        : "GNU-OpenMP",
+            "displayName" : "GNU-OpenMP",
+            "inherits"    : "default",
+            "cacheVariables" : {
+                "Kokkos_ENABLE_OPENMP" : {"type" : "BOOL"  , "value" : "ON"}
```
dalg24: I am not a big fan of the verbose syntax when we can get away with `"Kokkos_ENABLE_OPENMP": "ON"` but not blocking if others feel strongly about it.

kokkos.presets.json
```diff
@@ -0,0 +1,68 @@
...
+            "cacheVariables" : {
+                "CMAKE_EXPORT_COMPILE_COMMANDS" : {"type" : "BOOL"  , "value" : "ON"},
+                "CMAKE_BUILD_TYPE"              : {"type" : "STRING", "value" : "RELEASE"},
+                "CMAKE_CXX_STANDARD"            : {"type" : "STRING", "value" : "17"},
+                "CMAKE_CXX_EXTENSIONS"          : {"type" : "BOOL"  , "value" : "OFF"},
+                "BUILD_SHARED_LIBS"             : {"type" : "BOOL"  , "value" : "ON"}
+            }
+        },
+        {
+            "name"        : "GNU-OpenMP",
```
dalg24: Why the "GNU-" prefix?

dalg24 (commented):
* Definitely keep the old build, the standalone build without Kokkos should still be considered as the default way to build Kokkos Tools
* No objection to the CMake presets

crtrott:
@romintomasetti btw, if you want to get more formally involved with Kokkos Tools we are happy to do that :-)

kokkos.presets.json
romintomasetti: Indeed. Done.

kokkos.presets.json
romintomasetti: In this case, it seems that using `${sourceDir}/bin/nvcc_wrapper` is cleaner than relying on some environment variable.

kokkos.presets.json
romintomasetti: Indeed. I removed `displayName` :smile:

kokkos.presets.json
romintomasetti: I don't know. By default, this is `OFF` in `Kokkos`, see https://github.com/kokkos/kokkos/blob/55bbd9f175307a2ad42d205a3cebb5a32698545b/CMakeLists.txt#L225.

I don't have any opinion on that in this case. If you want, I can remove the `BUILD_SHARED_LIBS` parameter.

kokkos.presets.json
romintomasetti: I think the rationale is that if you compile with e.g. `clang` it would become `clang-OpenMP`. Fine to do what you prefer though.

romintomasetti:
> @romintomasetti btw, if you want to get more formally involved with Kokkos Tools we are happy to do that :-)

@crtrott I'd be more than happy to participate in the adventure!

kokkos.presets.json
```diff
@@ -0,0 +1,64 @@
...
+            "cacheVariables" : {
+                "Kokkos_ENABLE_OPENMP" : {"type" : "BOOL"  , "value" : "ON"}
+            }
+        },
+        {
+            "name"        : "GNU-Cuda",
+            "inherits"    : "default",
+            "cacheVariables" : {
+                "Kokkos_ENABLE_CUDA"        : {"type" : "BOOL"  , "value" : "ON"},
+                "Kokkos_ENABLE_CUDA_LAMBDA" : {"type" : "BOOL"  , "value" : "ON"},
```
dalg24: If not actually necessary for this PR I would recommend to drop since the option defaults to ON in 4.0 and there is discussion about deprecating it in 4.1 (see kokkos/kokkos#5964)

kokkos.presets.json
dalg24: I'd rather drop it and reintroduce later if necessary.

CMakePresets.json
```diff
@@ -0,0 +1,59 @@
...
+    "configurePresets" : [
+        {
+            "name"          : "default",
+            "binaryDir"     : "${sourceDir}/build-with-${presetName}",
+            "cacheVariables" : {
+                "CMAKE_EXPORT_COMPILE_COMMANDS"     : {"type" : "BOOL"  , "value" : "ON"},
+                "CMAKE_BUILD_TYPE"                  : {"type" : "STRING", "value" : "Release"},
+                "CMAKE_CXX_STANDARD"                : {"type" : "STRING", "value" : "17"},
+                "KokkosTools_REUSE_KOKKOS_COMPILER" : {"type" : "BOOL"  , "value" : "ON"},
+                "Kokkos_DIR"                        : {"type" : "STRING", "value" : "$penv{Kokkos_DIR}"}
```
dalg24: This is not the right option to help find Kokkos.  I'd expect one should use `Kokkos_ROOT`.
I suspect Kokkos is being found because CMake actually parses the environment for `<PackageName>_DIR`.
Either drop it or use Kokkos_ROOT.

dalg24 (commented):
I think we are getting close

kokkos.presets.json
romintomasetti: Agreed, and done :heavy_check_mark:

kokkos.presets.json
romintomasetti: Ok, removed! Thanks for the pointer ;)

CMakePresets.json
romintomasetti: I think I got inspiration from https://github.com/kokkos/kokkos-tools/blob/3855d8f6186b600b97ae81cd1d9417ad8e34c776/example/CMakeLists.txt#L1-L10, but you're right!

masterleinad: FWIW, it's totally fine to set `Kokkos_DIR` pointing to the directory holding `KokkosConfig.cmake`.

CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ include(cmake/configure_tpls.cmake)
 
 # Set policies
 cmake_policy(SET CMP0111 NEW) # error if library not found
+cmake_policy(SET CMP0074 NEW) # use <PackageName>_ROOT
```
dalg24:
```suggestion
```
No that is unnecessary with the cmake minimum required version >= 3.12

CMakePresets.json
```diff
@@ -0,0 +1,83 @@
...
+    "configurePresets" : [
+        {
+            "name"          : "default",
+            "binaryDir"     : "${sourceDir}/build-with-${presetName}",
+            "cacheVariables" : {
+                "CMAKE_EXPORT_COMPILE_COMMANDS"     : {"type" : "BOOL"  , "value" : "ON"},
+                "CMAKE_BUILD_TYPE"                  : {"type" : "STRING", "value" : "Release"},
+                "CMAKE_CXX_STANDARD"                : {"type" : "STRING", "value" : "17"},
+                "KokkosTools_REUSE_KOKKOS_COMPILER" : {"type" : "BOOL"  , "value" : "ON"},
+                "Kokkos_ROOT"                       : {"type" : "STRING", "value" : "$penv{Kokkos_ROOT}"}
```
dalg24: Not blocking but re-iterating that I don't think this is a legitimate use of $penv{...} I think $env{...} should be preferred.
Feel free to ignore but if you change make sure you tackle all occurrences.

CMakeLists.txt
romintomasetti: I was still having the warning in the CI, so I wanted to silence it. But fine if you want to remove the policy and accept the console warning ;)

CMakePresets.json
romintomasetti: Changed to `env`

romintomasetti:
@dalg24 Seems that there is an issue with the "space time stack" profiler... https://github.com/kokkos/kokkos-tools/actions/runs/4864236304/jobs/8672999895

Any idea?

### Edit

It seems `KokkosTools::SpaceTimeStack::kokkosp_init_library` is never called, and so the `global_state` is never initialized.

### Edit

I've added several prints in `Kokkos` and I think I found the issue.

In the function `Kokkos::Impl::OpenMPInternal::resize_thread_data`, that is called before the `KokkosTools::SpaceTimeStack::kokkosp_init_library` is called, it will do `ptr = space.allocate(alloc_bytes);`, that will try to call the space and time stack `KokkosTools::SpaceTimeStack::State::allocate` before `KokkosTools::SpaceTimeStack::kokkosp_init_library` is called.

```bash
Kokkos::Tools::Impl::InitializationStatus Kokkos::Tools::Impl::parse_environment_variables(Kokkos::Tools::InitArguments&)
void {anonymous}::initialize_internal(const Kokkos::InitializationSettings&)
void {anonymous}::initialize_internal(const Kokkos::InitializationSettings&): pre_initialize_internal
void {anonymous}::pre_initialize_internal(const Kokkos::InitializationSettings&)
void {anonymous}::initialize_internal(const Kokkos::InitializationSettings&): initialize_backends
void {anonymous}::initialize_backends(const Kokkos::InitializationSettings&): start
void Kokkos::Impl::ExecSpaceManager::initialize_spaces(const Kokkos::InitializationSettings&)
void Kokkos::Impl::ExecSpaceManager::initialize_spaces(const Kokkos::InitializationSettings&): 050_OpenMP
void Kokkos::Impl::OpenMPInternal::initialize(int): start
Kokkos::OpenMP::initialize WARNING: OMP_PROC_BIND environment variable not set
  In general, for best performance with OpenMP 4.0 or better set OMP_PROC_BIND=spread and OMP_PLACES=threads
  For best performance with OpenMP 3.1 set OMP_PROC_BIND=true
  For unit testing set OMP_PROC_BIND=false

void Kokkos::Impl::OpenMPInternal::initialize(int): instantiate OpenMP memory space
void Kokkos::Impl::OpenMPInternal::initialize(int): querying current max threads
void Kokkos::Impl::OpenMPInternal::initialize(int): current max threads 16
void Kokkos::Impl::OpenMPInternal::initialize(int): setting up thread local
void Kokkos::Impl::OpenMPInternal::initialize(int): getting OpenMPInternal singleton
void Kokkos::Impl::OpenMPInternal::initialize(int): resize thread data
void Kokkos::Impl::OpenMPInternal::resize_thread_data(size_t, size_t, size_t, size_t): start
void Kokkos::Impl::OpenMPInternal::resize_thread_data(size_t, size_t, size_t, size_t): rank 0 start
void Kokkos::Impl::OpenMPInternal::resize_thread_data(size_t, size_t, size_t, size_t): rank 0 allocate 28800
void* Kokkos::HostSpace::allocate(size_t) const: allocating 28800
void* Kokkos::HostSpace::allocate(const char*, size_t, size_t) const: allocating 28800
void* Kokkos::HostSpace::impl_allocate(const char*, size_t, size_t, Kokkos::Tools::SpaceHandle) const: allocating 28800
void* Kokkos::HostSpace::impl_allocate(const char*, size_t, size_t, Kokkos::Tools::SpaceHandle) const: profile library loaded so calling Kokkos::Profiling::allocateData(...)
void KokkosTools::SpaceTimeStack::State::allocate(KokkosTools::SpaceTimeStack::Space, const char*, const void*, uint64_t): allocating [unlabeled] in space 0(HOST)
[24f4e59e98fd:16590] *** Process received signal ***
[24f4e59e98fd:16590] Signal: Segmentation fault (11)
[24f4e59e98fd:16590] Signal code: Address not mapped (1)
[24f4e59e98fd:16590] Failing at address: 0xa0
```

### Edit

1. I guess all `Kokkos Tools` functions that rely on a global state should first verify that the global state is in a valid state before trying to call it.

2. I don't know how we should fix this.

crtrott:
This sounds like an issue on the Kokkos Core side, I don't think Kokkos should call any callback before the init thing is called

romintomasetti:
> This sounds like an issue on the Kokkos Core side, I don't think Kokkos should call any callback before the init thing is called

@crtrott Indeed. How do you think we should proceed to fix this? :ambulance:

masterleinad:
```diff
diff --git a/core/src/impl/Kokkos_Core.cpp b/core/src/impl/Kokkos_Core.cpp
index aec334261..a7f91203e 100644
--- a/core/src/impl/Kokkos_Core.cpp
+++ b/core/src/impl/Kokkos_Core.cpp
@@ -780,8 +780,14 @@ void post_initialize_internal(const Kokkos::InitializationSettings& settings) {
 }
 
 void initialize_internal(const Kokkos::InitializationSettings& settings) {
+  // Tools should only be used after the backends have been initialized (because
+  // they have not been initialized at that point) but if they have been set via
+  // set_callbacks they need to be present before calling post_initialize so
+  // that they get initialized as well.
+  Kokkos::Tools::Experimental::pause_tools();
   pre_initialize_internal(settings);
   initialize_backends(settings);
+  Kokkos::Tools::Experimental::resume_tools();
   post_initialize_internal(settings);
 }
```
should be enough to fix this.

romintomasetti:
@masterleinad Will you submit a PR?

Edit: Could the solution be that tools are initialized implicitly the first time they are called? Or should they simply be initialized at the very beginning of kokkos initialisation?

masterleinad:
> Edit: Could the solution be that tools are initialized implicitly the first time they are called? Or should they simply be initialized at the very beginning of kokkos initialisation?

That makes it even more difficult to argue about the initialization of `Kokkos Toosl`.
Instead of the `pause_tools`/`resume_tools` solution above we could also consider initializing the backends after the tools but that messes with the `Legion` use case and might therefore not be an option.

romintomasetti:
> > Edit: Could the solution be that tools are initialized implicitly the first time they are called? Or should they simply be initialized at the very beginning of kokkos initialisation?
> 
> That makes it even more difficult to argue about the initialization of `Kokkos Toosl`. Instead of the `pause_tools`/`resume_tools` solution above we could also consider initializing the backends after the tools but that messes with the `Legion` use case and might therefore not be an option.

What do you suggest then? Should we go for the `pause/resume` option for now and open an issue on both Kokkos and Kokkos Tools to try improve the initialization schemes?

masterleinad:
> What do you suggest then? Should we go for the `pause/resume` option for now and open an issue on both Kokkos and Kokkos Tools to try improve the initialization schemes?

See Kokkos/kokkos#6114.

romintomasetti:
I'm working on this again since https://github.com/kokkos/kokkos/pull/6114 has been merged :)

README.md
```diff
@@ -1,3 +1,7 @@
+![Standalone](https://github.com/kokkos/kokkos-tools/actions/workflows/build-standalone.yml/badge.svg)
+![Kokkos@develop](https://github.com/kokkos/kokkos-tools/actions/workflows/build-with-kokkos.yml/badge.svg)
+![clang-format](https://github.com/kokkos/kokkos-tools/actions/workflows/clang-format-check.yml/badge.svg)
+
```
dalg24: I am not a big fan of badges.  I would remove them but I am not blocking if others think otherwise.

.github/workflows/clang-format-check.yml
```diff
@@ -5,7 +5,7 @@ jobs:
     name: Formatting Check
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
```
dalg24, line 8: Next time please avoid unrelated changes.  (not asking you to revert)

CMakePresets.json
```diff
@@ -0,0 +1,86 @@
...
+    },
+    "configurePresets" : [
+        {
+            "name"          : "default",
+            "binaryDir"     : "${sourceDir}/build-with-${presetName}",
+            "cacheVariables" : {
+                "CMAKE_EXPORT_COMPILE_COMMANDS"     : {"type" : "BOOL"  , "value" : "ON"},
+                "CMAKE_BUILD_TYPE"                  : {"type" : "STRING", "value" : "Release"},
+                "CMAKE_CXX_STANDARD"                : {"type" : "STRING", "value" : "17"},
+                "KokkosTools_REUSE_KOKKOS_COMPILER" : {"type" : "BOOL"  , "value" : "ON"},
```
dalg24: Where is this used?

CMakePresets.json
```diff
@@ -0,0 +1,86 @@
...
+                "KokkosTools_ENABLE_EXAMPLES"       : {"type" : "BOOL"  , "value" : "ON"},
+                "KokkosTools_ENABLE_SINGLE"         : {"type" : "BOOL"  , "value" : "ON"},
+                "KokkosTools_ENABLE_MPI"            : {"type" : "BOOL"  , "value" : "ON"},
+                "Kokkos_ROOT"                       : {"type" : "STRING", "value" : "$env{Kokkos_ROOT}"}
```
dalg24:
```suggestion
```
This environment variable is used in the normal search procedure, why would we do that?

README.md
romintomasetti: No problem, I'll revert this ;) I was primarily just a mean to trigger the pipeline again (since https://github.com/kokkos/kokkos/pull/6114 was merged in the Kokkos develop branch, I wanted to check all would be fine).

.github/workflows/clang-format-check.yml
romintomasetti: Ok.

CMakePresets.json
romintomasetti: Sorry, it is unclear what you're pointing to. Are you referring to `KokkosTools_REUSE_KOKKOS_COMPILER`?

CMakePresets.json
romintomasetti: Indeed, weird that I also needed to repeat `Kokkos_ROOT` in the cache variables. Let's remove it and see :)

CMakePresets.json
dalg24: Yes

romintomasetti: Ok :)

The idea of the PR is to add a CI for joint `Kokkos` / `Kokkos Tools` builds. `KokkosTools_REUSE_KOKKOS_COMPILER` is a `Kokkos Tools` (advanced) option that forces to reuse the `Kokkos` compiler (e.g. `nvcc_wrapper` for Cuda I guess).

I thought it would be nice to ensure that during a joint build, both projects are on the same page for the compiler.

What's your opinion on that?

CMakePresets.json
```diff
@@ -0,0 +1,85 @@
+{
+    "version"   : 3,
+    "cmakeMinimumRequired" : {
+        "major" : 3,
+        "minor" : 22,
```
masterleinad: Why 22? Isn't the minimum we require 3.16?

kokkos.presets.json
```diff
@@ -0,0 +1,63 @@
+{
+    "version"   : 3,
+    "cmakeMinimumRequired" : {
+        "major" : 3,
+        "minor" : 22,
```
masterleinad: Again 3.16 is the minimum.

kokkos.presets.json
masterleinad: I think it's a good idea to only set required CMake options for `Kokkos` since that likely matches more closely a usual user configuration.

CMakePresets.json
romintomasetti: I removed `cmakeMinimumRequired` from the presets altogether :checkered_flag:

kokkos.presets.json
romintomasetti: I removed cmakeMinimumRequired from the presets altogether :checkered_flag:

CMakePresets.json
dalg24: Oh right.  I had grepped for it and must have misspelled it because now I can see it does exist.

kokkos.presets.json
romintomasetti: What do you want me to do exactly? Remove `BUILD_SHARED_LIBS`? Remove something else? :wink:

kokkos.presets.json
romintomasetti: Ok, I removed these `"type"` and `"value"` ;)

kokkos.presets.json
romintomasetti: @masterleinad

CMakePresets.json
romintomasetti: Is the current solution with `Kokkos_ROOT` OK for you both @masterleinad and @dalg24 ?

masterleinad (approved):
Looks OK to me.

kokkos.presets.json
```diff
@@ -0,0 +1,58 @@
...
+                "Kokkos_ARCH_VEGA906" : "ON",
+                "CMAKE_CXX_COMPILER"  : "hipcc"
+            }
+        }
+    ],
+    "buildPresets" : [
+        {
+            "name"                        : "OpenMP",
+            "configurePreset"             : "OpenMP",
+            "inheritConfigureEnvironment" : true
```
dalg24: Inheriting an empty environment is kinda pointless.

kokkos.presets.json
```diff
@@ -0,0 +1,58 @@
+{
+    "version"   : 3,
+    "configurePresets" : [
+        {
+            "name"          : "default",
+            "binaryDir"     : "${sourceDir}/build-with-${presetName}",
+            "cacheVariables" : {
+                "CMAKE_EXPORT_COMPILE_COMMANDS" : "ON",
```
dalg24: What is the point of exporting the compile command here?

CMakePresets.json
```diff
@@ -0,0 +1,80 @@
...
+    "version"   : 3,
+    "configurePresets" : [
+        {
+            "name"          : "default",
+            "binaryDir"     : "${sourceDir}/build-with-${presetName}",
+            "cacheVariables" : {
+                "CMAKE_EXPORT_COMPILE_COMMANDS"     : "ON",
+                "CMAKE_BUILD_TYPE"                  : "Release",
+                "CMAKE_CXX_STANDARD"                : "17",
+                "KokkosTools_REUSE_KOKKOS_COMPILER" : "ON",
```
dalg24: If you are going to use that "reuse cxx compiler option", what is the point of setting the cxx compiler below?

kokkos.presets.json
romintomasetti: Indeed, I'll remove that. I guess it was a leftover :smile:

kokkos.presets.json
romintomasetti: I guess it was needed mostly for me and myself in some early debugging phase. This isn't needed by default and you can consider it removed :smile:

CMakePresets.json
romintomasetti: As discussed, we are not even sure that `KokkosTools_REUSE_KOKKOS_COMPILER` works (as it does not set variables with `FORCE` for instance), and as we already set the compiler above it shouldn't be necessary.

CMakePresets.json
romintomasetti: I guess it's OK for you guys :wink:

dalg24 (approved):
Thank you

Issue #92, opened 2020-06-16
Documentation for writing a Kokkos connector?


tpatki:
@DavidPoliakoff 

@zfrye-llnl and I are thinking of writing a Variorum connector for Kokkos (https://github.com/LLNL/variorum). Is there any documentation on how to write one (or which functions: initialize/finalize, push/pop, begin/end_parallel_for are mandatory)?

DavidPoliakoff:
@tpatki 
Unfortunately, there's not great formal documentation on that, which I'll of course make a note of. Typical practice is to copy/paste something existing. As to what's required, only init and finalize are required. The full set of options is available here https://github.com/kokkos/kokkos-tools/wiki/Profiling-Hooks . I'm so happy you're taking a look at this, I'm excited to play with Variorum

DavidPoliakoff:
@zfrye-llnl : sorry that I missed you in the last message. You might also want to check out kokkos/kokkos#2422 , in which we're adding an autotuning extension to this interface. If it's feasible it might be interesting to develop a tool to minimize power without hurting performance, or things of that general flavor of optimization. That stuff is _utterly_ undocumented for the moment, so we'd need to collaborate, but I'd be happy to look at it with the two of you.

vlkale:
I would like to promote this github issue up as a priority for documentation of Kokkos Tools (using the new restructuredText format). 

Documentation of a Kokkos connector will be helpful for the ever-increasing vendor and open-source HPC tools available. In particular, this will be useful for autotuning tools available within the DoE along with new profiling and debugging tools available (in the last year) on leadership class supercomputers within the DoE.

vlkale:
Now part of the Kokkos Tools tutorial.

vlkale:
See the NERSC 2024 tutorial presentation on Kokkos to see the documentation / slides. I think the slides adequately provide information. 

If there is any other information anyone would like on documentation writing a Kokkos Tools connector, please feel free to request to reopen. 

Note this is an advanced topic and wouldn't be part of basic Kokkos Tools user documentation, which is a priority.

PR #190, opened 2023-04-27
FIx warnings in all tools built in the CI

### LINKED ISSUE #189: Squash warnings of llu print from Kokkos Tools
Remove warnings of llu print from Kokkos . This is seen in the latest CI for Kokkos Tools.

masterleinad:
Fixes #189.


vlkale:
Ah, thanks for going ahead and doing this! This LGTM and is how I was planning to fix.

vlkale:
@masterleinad Thanks again here. It's strange - this isn't passing the simple build CI now? It was just a squash of warnings right?

vlkale:
Ah sorry - I see this is a draft pull request. Perhaps there are some things in an intermediate state or still being worked on.

common/kokkos-sampler/kp_sampler_skip.cpp
```diff
@@ -238,7 +244,7 @@ extern "C" {
 
 namespace impl = KokkosTools::Sampler;
 
-EXPOSE_TOOL_SETTINGS(kokkosp_request_tool_settings)
+EXPOSE_TOOL_SETTINGS(impl::kokkosp_request_tool_settings)
```
masterleinad, line 247: This resulted in an infinite loop.

profiling/simple-kernel-timer/kp_json_writer.cpp
```diff
@@ -79,18 +79,8 @@ int main(int argc, char* argv[]) {
...
-
         int commandline_args = 1;
         while( (commandline_args<argc ) && (argv[commandline_args][0]=='-') ) {
-          if(strcmp(argv[commandline_args],"--delimiter")==0) {
-            delimiter=argv[++commandline_args][0];
-          }
-          if(strcmp(argv[commandline_args],"--fixed-width")==0) {
-            fixed_width=atoi(argv[++commandline_args]);
-          }
-
```
masterleinad, line 93: These command line arguments were parsed but never used. I'm not quite sure if we want to keep them or not.

dalg24: No that is a leftover from starting from kp_reader.  I think I wrote that guy and the delimiter of fixed width does not make sense for JSON

profiling/simple-kernel-timer/kp_json_writer.cpp
```diff
@@ -60,7 +60,7 @@ inline void write_json(std::ostream& os, KernelPerformanceInfo const& kp,
 int find_index(std::vector<KernelPerformanceInfo*>& kernels,
 	const char* kernelName) {
 
-	for(int i = 0; i < kernels.size(); i++) {
+	for(unsigned int i = 0; i < kernels.size(); i++) {
```
dalg24, line 63: I'd have gone for size_t but not blocking

common/kokkos-sampler/kp_sampler_skip.cpp
vlkale: @masterleinad Thanks for pointing this out! Do you have a particular code you tested this on that gave this result? 

By the way, I am currently creating test cases for all the tools connectors. (Also, from this, an idea I had is to create a generic tools utility and generic tools connector. This would address Kokkos Tools Github Issue #92 and make the development. ) This is a bigger effort but my focus is on the sampler. 

This fix you suggest should be part of a separate github PR.

masterleinad: No, I didn't actually test it. This just gave compiler warnings (for this to result in an infinite loop which makes sense) and that's why I fix it here.

vlkale: OK, got it.

That makes sense to put the fix in this PR.

Thanks.

crtrott (approved):

PR #192, opened 2023-05-02
Cleanup finding Kokkos in examples

dalg24:
CMake has native support for requiring a package to be found, there is no point on doing it ourselves.
Also I don't see why editing Kokkos_DIR is a good idea.  There is nothing wrong with pointing to the directory that contains the exported configuration and no obvious benefit to approximate the prefix where Kokkos was installed.  I suggest to drop the logic.
This was brought to may attention in https://github.com/kokkos/kokkos-tools/pull/188#discussion_r1182448051

### DIFF
--- a/example/CMakeLists.txt
+++ b/example/CMakeLists.txt
@@ -1,8 +1,5 @@
 # Find Kokkos
-find_package(Kokkos QUIET)
-if(NOT Kokkos_FOUND)
-  message(FATAL_ERROR "Kokkos not found, set Kokkos_ROOT properly (current Kokkos_ROOT=${Kokkos_ROOT})")
-endif()
+find_package(Kokkos REQUIRED)
 foreach(_i "1;2;3") # cut .../lib/cmake/Kokkos suffix
 cmake_path(GET Kokkos_DIR PARENT_PATH Kokkos_DIR)
 endforeach()
--- a/example/CMakeLists.txt
+++ b/example/CMakeLists.txt
@@ -1,8 +1,5 @@
 # Find Kokkos
 find_package(Kokkos REQUIRED)
-foreach(_i "1;2;3") # cut .../lib/cmake/Kokkos suffix
-cmake_path(GET Kokkos_DIR PARENT_PATH Kokkos_DIR)
-endforeach()
 message(STATUS "Found installed Kokkos at ${Kokkos_DIR}")
 
 # Create target executable


Rombur (approved):

example/CMakeLists.txt
```diff
@@ -1,11 +1,5 @@
 # Find Kokkos
-find_package(Kokkos QUIET)
-if(NOT Kokkos_FOUND)
-  message(FATAL_ERROR "Kokkos not found, set Kokkos_ROOT properly (current Kokkos_ROOT=${Kokkos_ROOT})")
-endif()
-foreach(_i "1;2;3") # cut .../lib/cmake/Kokkos suffix
-cmake_path(GET Kokkos_DIR PARENT_PATH Kokkos_DIR)
-endforeach()
```
masterleinad, line 8: This should have been
```
oreach(_i 1;2;3) # cut .../lib/cmake/Kokkos suffix
cmake_path(GET Kokkos_DIR PARENT_PATH Kokkos_DIR)
endforeach()
```
anyway. Note that the intention was to print the `Kokkos` root directory but we are now printing the path to the directory holding the configuration.

example/CMakeLists.txt
```diff
@@ -1,11 +1,5 @@
...
+find_package(Kokkos REQUIRED)
 message(STATUS "Found installed Kokkos at ${Kokkos_DIR}")
```
masterleinad, line 3: We already try to find `Kokkos` earlier and print the path there. I think duplicate that logic here instead of printing the path to the directory containing the configuration file
```suggestion
get_property(Kokkos_INSTALL_DIR TARGET Kokkos::kokkoscore PROPERTY INTERFACE_INCLUDE_DIRECTORIES)
cmake_path(GET Kokkos_INSTALL_DIR PARENT_PATH Kokkos_INSTALL_DIR)
message(STATUS "Found installed Kokkos at ${Kokkos_INSTALL_DIR}")
```

example/CMakeLists.txt
dalg24: See the description.  I understand what the intent was and I don't see any compelling reason to do so.

example/CMakeLists.txt
dalg24: I don't see an obvious benefit to do so and your suggestion add complexity.

crtrott (approved):
Fine with me

PR #193, opened 2023-05-02
Fix KokkosTools_ENABLE_SINGLE if systemtap is not available

masterleinad:
Follow-up to #183.


example/CMakeLists.txt
```diff
@@ -32,7 +32,9 @@ if(NOT WIN32)
   add_kp_test(memory_usage "memory-usage")
   add_kp_test(chrome_tracing "chrome-tracing")
   add_kp_test(space_time_stack "space-time-stack")
+if(KOKKOSTOOLS_HAS_SYSTEMTAP)
   add_kp_test(systemtap_connector "systemtap-connector")
+endif()
```
dalg24: Indentation is not right.

CMakeLists.txt
```diff
@@ -166,9 +164,13 @@ if(NOT WIN32 AND NOT APPLE)
   find_program(KOKKOSTOOLS_DTRACE_EXECUTABLE dtrace)
   if(KOKKOSTOOLS_DTRACE_EXECUTABLE)
     add_subdirectory(profiling/systemtap-connector)
+    set(KOKKOSTOOLS_HAS_SYSTEMTAP ON)
   else()
     message(STATUS "Skipping systemtap-connector (dtrace executable wasn't found)")
+    set(KOKKOSTOOLS_HAS_SYSTEMTAP OFF)
   endif()
+else()
+  set(KOKKOSTOOLS_HAS_SYSTEMTAP OFF)
```
dalg24, line 173: Why do you set to OFF?  I expect `#cmakedefine` will do the right thing if the variable is not defined at all.

masterleinad: I preferred to be explicit. I can change it if you feel strongly about it.

dalg24: I feel like it increases complexity but can live with it

dalg24 (approved):

crtrott (approved):

Issue #176, opened 2023-04-14
Fix Nvprof focused connector naming to nvtx, consistent with nvtx connector


vlkale:
Fix Nvprof focused connector naming to nvtx and associated cmakeLists.txt file, and ensure consistency in naming and functionality 

The nvprof-connector is already being renamed to nvtx-connector in #139. This PR is independent of changes to the nvtx connector renaming. 

NOTE: The KOKKOSTOOLS_HAS_NVPROF environment variable should probably be changed to KOKKOSTOOLS_HAS_NVTX as well. I am putting that aside for another PR.

vlkale:
@crtrott This is a fairly straightforward change (involving renaming). I have checked the files and filenames for naming consistency and that no other code outside these files is impacted. Can you please look at it quickly when possible? Thanks!

PR #202, opened 2023-08-24
KOKKOS_TOOLS_LIBS and kp_add_example for Kernel Filter

vlkale:
Kernel filter is used for nvtx_focused_connector and other tools. It is outdated given changes in user interface and cmake build systems. 
This PR is associated with #176 . 

This PR does the following: 

* Read the environment KOKKOS_TOOLS_LIBS in addition to KOKKOS_PROFILE_LIBRARY. 
* Use kp_add_example for Kernel Filter

### DIFF
--- a/common/kernel-filter/kp_kernel_filter.cpp
+++ b/common/kernel-filter/kp_kernel_filter.cpp
@@ -129,14 +129,16 @@ extern "C" void kokkosp_init_library(const int loadSeq,
       char* profileLibrary = getenv("KOKKOS_TOOLS_LIBS");
       if (NULL == profileLibrary) {  // check for backward compatibility with
                                      // old environment variable
-        printf(
-            "Checking KOKKOS_PROFILE_LIBRARY. WARNING: This is a deprecated "
-            "variable. Please use KOKKOS_TOOLS_LIBS.\n");
+        printf("KokkosP: KOKKOS_TOOLS_LIBS not set.\n");
         profileLibrary = getenv("KOKKOS_PROFILE_LIBRARY");
         if (NULL == profileLibrary) {
           printf("KokkosP: No library to call in %s\n", profileLibrary);
           exit(-1);
         }
+        else {
+         printf("KokkosP: Found that KOKKOS_PROFILE_LIBRARY is set and it will be used.\n");
+         printf("KokkosP: Note that KOKKOS_PROFILE_LIBRARY is a deprecated variable. " 
+         "Please use KOKKOS_TOOLS_LIBS in the future.\n");
       }  // end check for backward compatability
 
       char* envBuffer =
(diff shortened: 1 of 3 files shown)


common/kernel-filter/kp_kernel_filter.cpp
```diff
@@ -126,7 +126,19 @@ extern "C" void kokkosp_init_library(const int loadSeq,
            (filterKernels ? "enabled" : "disabled"));
 
     if (filterKernels) {
-      char* profileLibrary = getenv("KOKKOS_PROFILE_LIBRARY");
+      char* profileLibrary = getenv("KOKKOS_TOOLS_LIBS");
+      if (NULL == profileLibrary) {  // check for backward compatibility with
+                                     // old environment variable
+        printf(
+            "Checking KOKKOS_PROFILE_LIBRARY. WARNING: This is a deprecated "
+            "variable. Please use KOKKOS_TOOLS_LIBS.\n");
```
dalg24: Why emit this warning if `KOKKOS_PROFILE_LIBRARY ` was not actually set?

vlkale: Yeah, agreed - this is not a great warning text. It was copied and pasted from another library (nvtx-connector). The text after "WARNING: .." should be printed only in the case that KOKKOS_PROFILE_LIBRARY was set. I will change.

common/kernel-filter/kp_kernel_filter.cpp
```diff
@@ -182,7 +196,7 @@ extern "C" void kokkosp_init_library(const int loadSeq,
 
     printf("============================================================\n");
   }
-}
+}  // end kokkosp_init_library
```
dalg24, line 203: I don't really the point of marking the closing brace here.

common/kernel-filter/kp_kernel_filter.cpp
```diff
@@ -126,7 +126,21 @@ extern "C" void kokkosp_init_library(const int loadSeq,
...
+        profileLibrary = getenv("KOKKOS_PROFILE_LIBRARY");
+        if (NULL == profileLibrary) {
+          printf("KokkosP: No library to call in %s\n", profileLibrary);
+          exit(-1);
+        }
+        else {
+         printf("KokkosP: Found that KOKKOS_PROFILE_LIBRARY is set and it will be used.\n");
+         printf("KokkosP: Note that KOKKOS_PROFILE_LIBRARY is a deprecated variable. " 
+         "Please use KOKKOS_TOOLS_LIBS in the future.\n");
+      }  // end check for backward compatability
```
dalg24: Typo compat(i)bility.  I don't see the point of the comment. It is not even the right place.  You are missing a cling brace I don;'t even think this compiles

vlkale:
Note that this PR doesn't use the namespace for KokkosTools and Filter and kp_kernel_filter.cpp has functions with extern "C" . This will be fixed in a separate PR.

common/kernel-filter/kp_kernel_filter.cpp
vlkale: Thanks. I messed it up (in the process of trying to put that comment). See updated version .

crtrott (approved):

PR #211, opened 2023-10-04
cicd(gcc): add job with gcc 12 and nvcc 12.2.0

romintomasetti:
## Summary

* This PR adds a job using `Cuda 12.2.0` with `gcc 12`.
* It also fixes a missing `pthread.h` include in the `nvtx` connector (only from `gcc 12` onwards).

## Details

In order to make the missing `pthread.h` problem appear, I needed to compile with `gcc 12`, which is not the default compiler in `Ubuntu 22.04` images.

Therefore, the actions strategy matrix now has a `compiler` component, that can be set to `default` or anything else.

* `default` will install `gcc` and `g++`. For instance, in `Ubuntu 22.04` images it is version 11.
* Otherwise, use an array like `{c : my_c_compiler, cpp : my_cpp_compiler}`. It will install these compilers instead.

Then just set `CC` or `CXX` to instruct `CMake` which compilers to use.

Note that for the `Cuda` preset case, we also need to set `NVCC_WRAPPER_DEFAULT_COMPILER` because it will use `Kokkos`'s `nvcc_wrapper` that by default uses `g++` for the `c++` compiler.

### DIFF
--- a/.github/workflows/build-with-kokkos.yml
+++ b/.github/workflows/build-with-kokkos.yml
@@ -12,31 +12,69 @@ jobs:
         include:
           - image: ubuntu:22.04
             preset: OpenMP
+            compiler: default
           - image: nvidia/cuda:12.1.0-devel-ubuntu22.04
             preset: Cuda
+            compiler: default
+          - image: nvidia/cuda:12.2.0-devel-ubuntu22.04
+            preset: Cuda
+            compiler: {cpp: g++-12, c: gcc-12}
           - image: rocm/dev-ubuntu-22.04:5.4
             preset: ROCm
+            compiler: default
           - image: rocm/dev-ubuntu-22.04:5.7
             preset: ROCm
+            compiler: default
     container:
       image: ${{ matrix.image }}
     env:
       Kokkos_ROOT: /opt/kokkos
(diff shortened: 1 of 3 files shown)


romintomasetti:
Hi @masterleinad @vlkale ! Would you have a moment to review this one? Thanks!

vlkale (approved):
LGTM. Thanks!

masterleinad (commented):
Why do you want to test with both `cuda-12.1.0` and `cuda-12.2.0` and not some older version (e.g., the oldest one `Kokkos` supports)?

romintomasetti:
> Why do you want to test with both `cuda-12.1.0` and `cuda-12.2.0` and not some older version (e.g., the oldest one `Kokkos` supports)?

Because that's our use case. We strive to be up-to-date with latest versions. In our experience, this helps detecting bugs.

If you want, I can make another PR that changes `cuda-12.1.0` to an older version :wink:

crtrott:
Yeah lets change the 12.1 to an old version. I am ok with having two CUDA builds in here: one the newest and one if possible the oldest Kokkos supports (right now 11.0)

Issue #191, opened 2023-04-28
Create unit tests for all Kokkos Tools libraries and utilities


vlkale:
Complementing the example directory, a set of unit tests and a framework for testing needs to be created for all Kokkos Tools libraries. This includes in-house tooling libraries, e.g., kernel-logger, vendor connectors, e.g., roctx-connector, and utilities, e.g., Kokkos-sampler, and sophisticated tpls, apex Kokkos Tool library.

In the near-term and of priority: the focus will be on new tools integrated being into the Kokkos Tools repo, in particular the sampler utility and fixes in the nvtx-connnector, which is expected to have heavy use. (The rocprof-connector and vtune-connector should also have similar tests). 

Longer-term: more tests, perhaps in part community-driven, will be incorporated in to the set of tests.

This may eventually involve a testing framework like Google Test and the testing would then happen within the new Kokkos Tools make build system and  nightly CI builds  of Kokkos Tools

vlkale:
I propose a process for developing these tests that is directly motivated by Kokkos applications using Kokkos Tools.


For now, let's define a unit as callback of a particular tool. Each unit test should verify (is the answer correct given the function's implementation), validate (is the answer correct based on requirements of function defined), and assess performance. These three aspects give the Kokkos Tools users an assessment of quality of the implementation. 

The process for establishing a set of unit tests for the Kokkos Tools would be something like: 

----

**1st phase**: Obtain simple example codes from applications and benchmarks, and ensure they work.  An example of a test of the sampler utility's parallel_for tools callback applied for the simple kernel timer is shown in the file. [outputForSampler with Atomic Benchmark](https://github.com/kokkos/kokkos-tools/files/11409469/outputForSampler-Kto-timers-AtomicBenchmark.pdf)

**2nd phase**: By stripping out functions and code from applications: Create tests of output of tool, with one test per unit, via assertions in code. 

**3rd phase**: Use testing framework like ctest (and could be subsequently using Google Test framework) and integrate the tests under the framework into the Kokkos Tools cmake build system.

----

The 1st and 2nd phase would be stored in a separate repo and be not part of the Kokkos Tools repo. The code developed in the 3rd phase would be the official tests of the repo, which would be community-built and agreed upon via a community of Kokkos users.  

This process could be followed or follow that of tests in Kokkos-core.

PR #218, opened 2023-11-06
tool(space-time-stack): demangle name

romintomasetti:
## Summary

This PR ensures that the tool `space-time-stack` demangles kernel names.

## Side quests, questions and related issues

- [ ] I created new files to store what's common across different tools. Tell me what you think @vlkale @dalg24 @masterleinad and how this should be done in your opinion. (`SpaceHandle.hpp`, `utils_demangle.hpp` and so on) I think this touches upon the more general problem of duplication in the tools and how this should be tackled.
- [x] ~~I changed `Kokkos_Tools_OptimzationGoal` to `Kokkos_Tools_OptimizationGoal` because I think it's a typo. I guess @dalg24 will agree that this deserves a separate PR. Note that the same change will be needed in `Kokkos`.~~ Discussion moved to #221.
- [ ] I started targeting #191. I guess I'll need *Google Test* to do so, as mentioned in the issue by @vlkale. I suggest that we add it as a TPL through a submodule approach.
- [ ] For easy testing, I suggest that things like `struct Stack` defined in `kp_space_time_stack.cpp` has a declaration in a corresponding header file, so tests can directly go ask such structures what they captured during a test. Tell me what you think.
- [ ] I was really surprised that the `SpaceHandle` recognizes the space type by looking at the first letters of its `name` attribute. Is there any reason why it's done like this and not *via* an enumeration? To me, this seems inefficient, because in *e.g.* `Kokkos` you'll need to create the `SpaceHandle` by creating a string with the space name, and pass it to the tools, that will need to parse that string. This seems to have more impact on the runtime that simply passing an `enum`... Moreover, several tools define their own enumeration for a subset of available `Kokkos` spaces. I would propose that by default the `SpaceHandle` only has an attribute `type` that comes from an enumeration like the one I defined in `SpaceHandle.hpp`. What do you think?
- [ ] I created a new `enum FrameType` that seeks to homogenize how kernel types are treated amongst tools, and how they are printed. This also deserves another PR if you agree on the principle.

### DIFF
--- a/common/FrameType.hpp
+++ b/common/FrameType.hpp
@@ -1,0 +1,53 @@
+#ifndef KOKKOSTOOLS_COMMON_FRAMETYPE_HPP
+#define KOKKOSTOOLS_COMMON_FRAMETYPE_HPP
+
+#include <ostream>
+#include <string>
+
+namespace KokkosTools {
+
+enum FrameType {
+    PARALLEL_FOR    = 0,
+    PARALLEL_REDUCE = 1,
+    PARALLEL_SCAN   = 2,
+    REGION          = 3,
+    COPY            = 4
+};
+
+//! A @ref FrameType::REGION is not a kernel.
+inline constexpr bool is_a_kernel(const FrameType type) {
+  switch (type) {
+    case PARALLEL_FOR   : return true;
+    case PARALLEL_REDUCE: return true;
+    case PARALLEL_SCAN  : return true;
+    case REGION         : return false;
+    case COPY           : return true;
+    default: throw type;
+  }
+}
+
+inline std::string to_string(const FrameType t) {
+  switch (t) {
+    case PARALLEL_FOR   : return "PARALLEL_FOR";
+    case PARALLEL_REDUCE: return "PARALLEL_REDUCE";
+    case PARALLEL_SCAN  : return "PARALLEL_SCAN";
+    case REGION         : return "REGION";
+    case COPY           : return "COPY";
+    default: throw t;
+  }
+}
+
+inline std::ostream &operator<<(std::ostream &os, const FrameType kind) {
+  switch (kind) {
+    case PARALLEL_FOR   : os << "[for]"; break;
+    case PARALLEL_REDUCE: os << "[reduce]"; break;
+    case PARALLEL_SCAN  : os << "[scan]"; break;
+    case REGION         : os << "[region]"; break;
+    case COPY           : os << "[copy]"; break;
+  };
+  return os;
+}
+
+} // namespace KokkosTools
+
+#endif // KOKKOSTOOLS_COMMON_FRAMETYPE_HPP
--- a/common/SpaceHandle.hpp
+++ b/common/SpaceHandle.hpp
@@ -1,0 +1,67 @@
+#ifndef KOKKOSTOOLS_COMMON_SPACEHANDLE_HPP
+#define KOKKOSTOOLS_COMMON_SPACEHANDLE_HPP
+
+#include "impl/Kokkos_Profiling_C_Interface.h"
+
+#include <ostream>
+#include <string>
+
+namespace KokkosTools
+{
+//! A @c Kokkos space type has a unique name.
+using SpaceHandle = Kokkos_Profiling_SpaceHandle;
+
+//! Supported @c Kokkos spaces.
+enum Space {
+    HOST = 0,
+    CUDA = 1,
+    HIP  = 2,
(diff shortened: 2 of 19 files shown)


vlkale:
@maartenarnst  Thanks for this earlier. This PR is helpful. Sorry for the delayed reply due to conferences and US holidays this time of year. 


> ## Summary
> This PR ensures that the tool `space-time-stack` demangles kernel names.
> 
> ## Side quests, questions and related issues
> * [ ]  I created new files to store what's common across different tools. Tell me what you think @vlkale @dalg24 @masterleinad and how this should be done in your opinion. (`SpaceHandle.hpp`, `utils_demangle.hpp` and so on) I think this touches upon the more general problem of duplication in the tools and how this should be tackled.


Yes, the issue of duplication across tools will definitely make development of tool connector much easier for someone new to Kokkos Tools. Perhaps it is OK  

> * [ ]  I changed `Kokkos_Tools_OptimzationGoal` to `Kokkos_Tools_OptimizationGoal` because I think it's a typo. I guess @dalg24 will agree that this deserves a separate PR. Note that the same change will be needed in `Kokkos`.

Thanks for that! I don't know how much this is being used and how much it impacts other Kokkos Tools from a quick look. Yes, let me know if you can easily submit a new PR on your end, and we will look at it to get this quick fix merged. 

> * [ ]  I started targeting [Create unit tests for all Kokkos Tools connectors and utilities #191](https://github.com/kokkos/kokkos-tools/issues/191). I guess I'll need _Google Test_ to do so, as mentioned in the issue by @vlkale. I suggest that we add it as a TPL through a submodule approach.

Yes, I think Google Test was a generally accepted way to do testing of Kokkos Tools when I wrote that Ghub Issue, having discussed with @crtrott previously months ago. Also, Google Tests is  used in Kokkos Core.  If you know of another dominant and here-to-stay framework for unit tests, let us know, but my opinion from experience with GTests is that it is straightforward and reasonably documented and still the best way to go for now. 

> * [ ]  For easy testing, I suggest that things like `struct Stack` defined in `kp_space_time_stack.cpp` has a declaration in a corresponding header file, so tests can directly go ask such structures what they captured during a test. Tell me what you think.

I think that this is a good idea. I don't see any issues with putting struct Stack in the header file. What other parts of kp_space_time_stack.cpp were you thinking to move? 

> * [ ]  I was really surprised that the `SpaceHandle` recognizes the space type by looking at the first letters of its `name` attribute. Is there any reason why it's done like this and not _via_ an enumeration? To me, this seems inefficient, because in _e.g._ `Kokkos` you'll need to create the `SpaceHandle` by creating a string with the space name, and pass it to the tools, that will need to parse that string. This seems to have more impact on the runtime that simply passing an `enum`... Moreover, several tools define their own enumeration for a subset of available `Kokkos` spaces. I would propose that by default the `SpaceHandle` only has an attribute `type` that comes from an enumeration like the one I defined in `SpaceHandle.hpp`. What do you think?

Deferring to @dalg24, @crtrott and maybe @masterleinad. Your points seem reasonable and good to me but we probably ought to think whether what you suggest is the only/best solution.

> * [ ]  I created a new `enum FrameType` that seeks to homogenize how kernel types are treated amongst tools, and how they are printed. This also deserves another PR if you agree on the principle.

Deferring to @dalg24, @crtrott.  I like the motivation of the problem. We probably need more discussion on the solution to the problem. 

For the last two bullet points, I suggest putting up a GitHub Issue for each first in order to identify the right solutions for them.

romintomasetti:
@vlkale I've made #234 for discussing about the `enum FrameType`.

vlkale:
I have already mentioned this to @romintomasetti through DM but I am passing on here that there is a problem with Google Test relying on std::cout. A large number of Kokkos Tools connectors use printf's and not std::cout, including kernel-logger. 

@romintomasetti I am thinking we will probably want to go with the Python script post-processing solution right now, and let all tools connectors continue to use printf's.

vlkale:
> Screenshot 2024-03-25 at 12 59 42 PM I have already mentioned this to @romintomasetti through DM but I am passing on here that there is a problem with Google Test relying on std::cout
> @romintomasetti I am thinking we will probably want to go with the Python script post-processing solution right now, and let all tools connectors continue to use printf's.

Also, the above screenshot was for the Kokkos Serial backend on my 2022 MacOS, and the output shown is for test_space_time_stack_demangling. 

The problem is likely 'output.str()' is " " , thanks to @masterleinad 's suggestion.

PR #219, opened 2023-11-13
Use eventset in roctxconnector (as in nvtxconnector)

maartenarnst:
This PR 
- updates `roctxconnector` by defining all the functions with the event set and the helper macros, like what is done in `nvtxconnector`. One of the interests of this approach is that it allows the callbacks to be set in a Kokkos program with `get_event_set("roctx-connector",...)` and then `set_callbacks`.

On the side, this PR also
- introduces the `namespace KokkosTools::ROCTXConnector`
- updates `kokkosp_request_tool_settings` such that fencing can be controlled with `KOKKOS_TOOLS_GLOBALFENCES`

Note that I've also tried to update the `CMakeLists.txt` for the roctxconnector to see if something could be done with a CMake target for roctx/roctracer, but I wasn't able to find such targets in the ROCm directories.

This is joint work with @romintomasetti.

@vlkale, @crtrott, would you have a moment to take a look, or refer to someone to take a look? Thanks in advance!


vlkale:
@maartenarnst @romintomasetti  Thanks very much for this. I plan to review this later this week or early next.

From looking at your PR code changes, it makes sense to me and would be useful change.

profiling/nvtx-connector/Makefile
```diff
@@ -1,5 +1,5 @@
 CXX=g++
-CXXFLAGS=-O3 -std=c++11 -g -I$(CUDA_ROOT)/include
+CXXFLAGS=-O3 -std=c++11 -g -I$(CUDA_ROOT)/include/
```
dalg24: Why did you change this?  In any case it does not seem to belong to this PR.

profiling/roctx-connector/CMakeLists.txt
```diff
@@ -3,5 +3,5 @@ find_path(ROCM_ROCTX_INCLUDE roctx.h REQUIRED HINTS $ENV{ROCM_PATH}/include/roct
 
 kp_add_library(kp_roctx_connector kp_roctx_connector.cpp)
 
-target_include_directories(kp_roctx_connector PRIVATE ${ROCM_ROCTX_INCLUDE})
-target_link_libraries(kp_roctx_connector PRIVATE ${ROCM_ROCTX_LIB})
+target_include_directories(kp_roctx_connector PUBLIC ${ROCM_ROCTX_INCLUDE})
+target_link_libraries(kp_roctx_connector PUBLIC ${ROCM_ROCTX_LIB})
```
dalg24: Why did you change?
I can see how the linking would need some symbols to be private but not why the include directories would be needed downstream.

profiling/roctx-connector/kp_roctx_connector.cpp
```diff
@@ -29,29 +31,40 @@ struct Section {
...
+
+  const char* tool_global_fences = std::getenv("KOKKOS_TOOLS_GLOBALFENCES");
+  if (tool_global_fences) {
+    tool_globfences = (atoi(tool_global_fences) != 0);
+  }
+
   std::cout << "-----------------------------------------------------------\n"
             << "KokkosP: ROC Tracer Connector (sequence is " << loadSeq
             << ", version: " << interfaceVer << ")\n"
+            << "Global fences: " << (tool_globfences ? "ON" : "OFF") << "\n"
```
dalg24: I feel like this is something one would only want to see in verbose mode.

profiling/roctx-connector/kp_roctx_connector.cpp
```diff
@@ -61,66 +74,120 @@ KokkosP: Finalization of ROC Tracer Connector. Complete.
...
   *fID = roctxRangeStart(name);
 }
 
-extern "C" void kokkosp_end_fence(const uint64_t fID) { roctxRangeStop(fID); }
+void kokkosp_end_fence(const uint64_t fID) { roctxRangeStop(fID); }
+
+Kokkos::Tools::Experimental::EventSet get_event_set() {
+  Kokkos::Tools::Experimental::EventSet my_event_set;
+  memset(&my_event_set, 0,
+         sizeof(my_event_set));  // zero any pointers not set here
```
dalg24, line 131: Uggg.  That thing really something that should be handled at construction.
I suspect value-initialization
```C++
EventSet my_event_set{};
```
does what you want.

profiling/nvtx-connector/Makefile
maartenarnst: It's a leftover from when I mistakenly changed something in the `nvtx-connector`. I've reverted this change.

profiling/roctx-connector/CMakeLists.txt
maartenarnst: These are leftovers from when I was trying to make the `roctx-connector` work with `rocprof`. I've reverted the change.

In fact, in my first tries, I used `rocprof` as follows:

```
rocprof --hip-trace --hsa-trace --roctx-trace ./a.out --kokkos-tools-libs=${KOKKOS_TOOLS_DIR}/lib/libkp_roctx_connector.so
``` 

That didn't work, in the sense that even though this command asks `rocprof` to use `--roctx-trace`, `rocprof` didn't include the markers in the `results.json` trace. I suspect that the reason may be that `rocprof` wants to see the executable linked to `roctx64.so`. That may not be the case when the connector is linked at runtime via dlsym. And so these leftovers came from attempts to make `roctx64.so` appear among the linked libraries.

- https://github.com/ROCm-Developer-Tools/rocprofiler/blob/8222e98dced7581c8e1e0666457a1a78fcf8f306/src/core/session/tracer/src/loader.h#L40-L58

profiling/roctx-connector/kp_roctx_connector.cpp
maartenarnst: Indeed. It seems there's currently no "verbose" bode in the different connectors. So, I've reverted the change.

profiling/roctx-connector/kp_roctx_connector.cpp
maartenarnst: Indeed :) I used this piece of code because all connectors are currently doing it this way. It may be best to address the issue for all connectors in a separate PR? One way forward might be to provide default initialisers using `= NULL` for the function pointers in the declaration of the struct in `Kokkos_Profiling_C_interface.h`?

dalg24: > I used this piece of code because all connectors are currently doing it this way.

You are right.  This is fine then.

And yes definitely fix elsewhere.

dalg24 (dismissed):

profiling/roctx-connector/kp_roctx_connector.cpp
maartenarnst: OK. I made issue 

- https://github.com/kokkos/kokkos-tools/issues/220

maartenarnst:
Hi @dalg24. Indeed, I've just improved the formatting. It seems the workflows need an approval to run. Would you have a moment to give an approval to see if it passes the formatting test now? Thanks.

masterleinad:
> Hi @dalg24. Indeed, I've just improved the formatting. It seems the workflows need an approval to run. Would you have a moment to give an approval to see if it passes the formatting test now? Thanks.

@vlkale

dalg24 (approved):

vlkale:
> > Hi @dalg24. Indeed, I've just improved the formatting. It seems the workflows need an approval to run. Would you have a moment to give an approval to see if it passes the formatting test now? Thanks.
> 
> 
> 
> @vlkale 

@dalg24 cc @masterleinad thanks

PR #222, opened 2023-12-04
papi(bug): use `kp_add_library` to ensure it is installed

romintomasetti:
The `PAPI` connector was not part of the installed files, because it was added with `add_library` instead of `kp_add_library`.

This PR also adds `libpapi-dev` in the `build-with-kokkos.yaml` pipeline so we can at least build the `PAPI` connector.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -9,7 +9,8 @@
                 "CMAKE_CXX_STANDARD"                : "17",
                 "KokkosTools_ENABLE_EXAMPLES"       : "ON",
                 "KokkosTools_ENABLE_SINGLE"         : "ON",
-                "KokkosTools_ENABLE_MPI"            : "ON"
+                "KokkosTools_ENABLE_MPI"            : "ON",
+                "KokkosTools_ENABLE_PAPI"           : "ON"
             }
         },
         {
--- a/.github/workflows/build-with-kokkos.yml
+++ b/.github/workflows/build-with-kokkos.yml
@@ -71,12 +71,13 @@ jobs:
                   exit -1
           esac
 
-      - name: Install CMake, OpenMPI and dtrace
+      - name: Install CMake, OpenMPI, PAPI and dtrace
         run: |
           apt --yes --no-install-recommends install \
             cmake make \
             libopenmpi-dev \
-            systemtap-sdt-dev
+            systemtap-sdt-dev \
+            libpapi-dev
       - name: Compile and install Kokkos
         working-directory: kokkos
         run: |
--- a/profiling/papi-connector/CMakeLists.txt
+++ b/profiling/papi-connector/CMakeLists.txt
@@ -1,3 +1,3 @@
-add_library(kp_papi_connector SHARED kp_papi_connector.cpp)
+kp_add_library(kp_papi_connector kp_papi_connector.cpp)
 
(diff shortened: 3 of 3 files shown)


romintomasetti:
@vlkale @dalg24

CMakePresets.json
```diff
@@ -9,7 +9,8 @@
                 "CMAKE_CXX_STANDARD"                : "17",
                 "KokkosTools_ENABLE_EXAMPLES"       : "ON",
                 "KokkosTools_ENABLE_SINGLE"         : "ON",
-                "KokkosTools_ENABLE_MPI"            : "ON"
+                "KokkosTools_ENABLE_MPI"            : "ON",
+                "KokkosTools_ENABLE_PAPI"           : "ON"
```
dalg24, line 13: I understand that is the easiest way to get this tested in all builds but I am not sure whether enabling PAPI should be configured as the "default" build.  That is the name of the preset that you changed.

dalg24 (approved):
Maybe we should not get hung up on the name and revisit later if the need arise.

CMakePresets.json
romintomasetti: At first, I thought it would only make sense to test the `PAPI` connector for CPU-related execution spaces.

But according to https://github.com/icl-utk-edu/papi/wiki (thought I haven't read the details), `PAPI` claims to also work for GPUs (*e.g.* https://github.com/icl-utk-edu/papi/wiki/PAPI-Component-ROCM).

I'm not against testing `PAPI` only for the presets like `OpenMP`. I'll do what you advise :wink:

dalg24: (I did not request changes)

vlkale: > At first, I thought it would only make sense to test the `PAPI` connector for CPU-related execution spaces.
> 
> 
> 
> But according to https://github.com/icl-utk-edu/papi/wiki (thought I haven't read the details), `PAPI` claims to also work for GPUs (*e.g.* https://github.com/icl-utk-edu/papi/wiki/PAPI-Component-ROCM).
> 
> 
> 
> I'm not against testing `PAPI` only for the presets like `OpenMP`. I'll do what you advise :wink:

Thanks very much for these changes and thanks for your particular question on PAPI GPU library support.

I absolutely think PAPI should be tested for more than presets like OpenMP. 


Given my own prior experience with PAPI, and recent interactions with Kokkos users and developers, Kokkos should allow for PAPI profiling for GPUs and not just CPUs. With support for PAPI in CUDA for more than a few years and recent maturity of PAPI for HIP over the past year, I think it is fundamentally important that Kokkos Tools do tests on PAPI GPU libraries.

Broadly speaking: Kokkos users should not have to rely on vendor's derived metrics through their visualization tools. They should be able to have direct access to hardware counters. I think this true for any scientific application program using CUDA, HIP, SyCL. 

I don't know if there is a SYCL PAPI.



With all the above said, I think that In principle, putting this particular change on PAPI GPU libraries in a separate Kokkos Tools PR may make more sense so that your non PAPI GPU related can be merged quickly.

 @dalg24 ?

vlkale:
> Maybe we should not get hung up on the name and revisit later if the need arise.

Agreed.

vlkale:
Also, note that PAPI isn't available everywhere, sometimes even on CPUs (so I think having it as the default Kokkos Tools build may not be desirable).

masterleinad (approved):

PR #225, opened 2023-12-21
space-time-stack: allow demangled names

romintomasetti:
As promised in #218, this PR extracts the part relevant to demangling names.

1. The demangling function is now shared in `common/utils/demangle.hpp`.
2. The *space time stack* tool now demangles the names.

This is how the report looks like with the changes:
```bash
9: <average time> <percent of total time> <percent time in Kokkos> <percent MPI imbalance> <remainder> <kernels per second> <number of calls> <name> [type]
9: ===================
9: |-> 1.49e-02 sec 43.8% 100.0% 0.0% ------ 1 Tester<Kokkos::OpenMP, 5ul> [for]
9: |-> 1.48e-02 sec 43.5% 100.0% 0.0% ------ 2 named kernel [for]
9: |-> 4.15e-03 sec 12.2% 100.0% 0.0% ------ 1 Tester<Kokkos::OpenMP, 5ul>/Tester<Kokkos::OpenMP, 5ul>::TagUnnamed [for]
```
and how it looks without the changes:
```bash
<average time> <percent of total time> <percent time in Kokkos> <percent MPI imbalance> <remainder> <kernels per second> <number of calls> <name> [type]
===================
|-> 1.79e-02 sec 59.4% 100.0% 0.0% ------ 2 named kernel [for]
|-> 7.90e-03 sec 26.2% 100.0% 0.0% ------ 1 6TesterIN6Kokkos6OpenMPELm5EE [for]
|-> 4.20e-03 sec 13.9% 100.0% 0.0% ------ 1 6TesterIN6Kokkos6OpenMPELm5EE/N6TesterIN6Kokkos6OpenMPELm5EE10TagUnnamedE [for]
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,6 +45,8 @@ option(KokkosTools_ENABLE_MPI     "Enable MPI support"              OFF)
 option(KokkosTools_ENABLE_CALIPER "Enable building Caliper library" OFF)
 option(KokkosTools_ENABLE_APEX    "Enable building Apex library"    OFF)
 option(KokkosTools_ENABLE_EXAMPLES "Build examples"                 OFF)
+option(KokkosTools_ENABLE_TESTS    "Build tests"                    OFF)
+
 # Advanced settings
 option(KokkosTools_REUSE_KOKKOS_COMPILER "Set the compiler and flags based on installed Kokkos settings" OFF)
 mark_as_advanced(KokkosTools_REUSE_KOKKOS_COMPILER)
@@ -115,6 +117,9 @@ include_directories(${CMAKE_CURRENT_SOURCE_DIR}/profiling/all)
 
 set(COMMON_HEADERS_PATH ${CMAKE_CURRENT_BINARY_DIR}/common)
 include_directories(${COMMON_HEADERS_PATH})
+
+# Allow all tools to include any file.
+include_directories(${CMAKE_CURRENT_SOURCE_DIR}/common)
 
 set(SINGLELIB_PROFILERS "" CACHE STRING "" FORCE)
 
@@ -262,6 +267,12 @@ if(KokkosTools_ENABLE_EXAMPLES)
     enable_testing()
     add_subdirectory(example)
   endif()
+endif()
+
+# Tests
+if(KokkosTools_ENABLE_TESTS)
+  enable_testing()
+  add_subdirectory(tests)
 endif()
 
 # Install exports
--- a/common/utils/demangle.hpp
+++ b/common/utils/demangle.hpp
@@ -1,0 +1,74 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#ifndef KOKKOSTOOLS_COMMON_UTILS_DEMANGLE_HPP
+#define KOKKOSTOOLS_COMMON_UTILS_DEMANGLE_HPP
+
+#include <string>
+
+#if defined(__GXX_ABI_VERSION)
+#define HAVE_GCC_ABI_DEMANGLE
+#endif
+
+#if defined(HAVE_GCC_ABI_DEMANGLE)
+#include <cxxabi.h>
+#endif  // HAVE_GCC_ABI_DEMANGLE
+
+namespace KokkosTools {
+
+//! Demangle @p mangled_name.
+inline std::string demangleName(const std::string_view mangled_name) {
+#if defined(HAVE_GCC_ABI_DEMANGLE)
+  int status = 0;
+
+  char* demangled_name =
+      abi::__cxa_demangle(mangled_name.data(), nullptr, nullptr, &status);
+
+  if (demangled_name) {
+    std::string ret(demangled_name);
+    std::free(demangled_name);
+    return ret;
+  }
+#endif
+  return std::string(mangled_name);
+}
+
+/**
+ * @brief Demangle @p mangled_name.
+ *
+ * This function supports @c Kokkos convention from
+ * @c Kokkos::Impl::ParallelConstructName.
+ *
+ * For instance, a kernel launched with a tag would appear as
+ * "<functor type>/<tag type>".
+ */
+inline std::string demangleNameKokkos(const std::string_view mangled_name) {
+  if (size_t pos = mangled_name.find('/', 0);
+      pos != std::string_view::npos && pos > 0) {
+    /// An explicit copy of the first part of the string is needed, because
+    /// @c abi::__cxa_demangle will parse the pointer until its NULL-terminated.
+    return demangleName(std::string(mangled_name.substr(0, pos)))
+        .append("/")
+        .append(
+            demangleName(mangled_name.substr(pos + 1, mangled_name.size())));
+  } else {
+    return demangleName(mangled_name);
+  }
+}
+
+}  // namespace KokkosTools
+
+#endif  // KOKKOSTOOLS_COMMON_UTILS_DEMANGLE_HPP
--- a/profiling/simple-kernel-timer/kp_json_writer.cpp
+++ b/profiling/simple-kernel-timer/kp_json_writer.cpp
@@ -83,7 +83,7 @@ int main(int argc, char* argv[]) {
       KernelPerformanceInfo* new_kernel =
           new KernelPerformanceInfo("", PARALLEL_FOR);
       if (new_kernel->readFromFile(the_file)) {
-        if (strlen(new_kernel->getName()) > 0) {
+        if (!new_kernel->getName().empty()) {
           int kernelIndex = find_index(kernelInfo, new_kernel->getName());
 
           if (kernelIndex > -1) {
--- a/profiling/simple-kernel-timer/kp_reader.cpp
+++ b/profiling/simple-kernel-timer/kp_reader.cpp
@@ -64,7 +64,7 @@ int main(int argc, char* argv[]) {
       KernelPerformanceInfo* new_kernel =
           new KernelPerformanceInfo("", PARALLEL_FOR);
       if (new_kernel->readFromFile(the_file)) {
-        if (strlen(new_kernel->getName()) > 0) {
+        if (!new_kernel->getName().empty()) {
           int kernelIndex = find_index(kernelInfo, new_kernel->getName());
 
           if (kernelIndex > -1) {
@@ -104,7 +104,7 @@ int main(int argc, char* argv[]) {
     if (kernelInfo[i]->getKernelType() != REGION) continue;
     if (fixed_width)
       printf("- %100s\n%11s%c%15.5f%c%12" PRIu64 "%c%15.5f%c%7.3f%c%7.3f\n",
-             kernelInfo[i]->getName(),
+             kernelInfo[i]->getName().c_str(),
              (kernelInfo[i]->getKernelType() == PARALLEL_FOR)
                  ? (" (ParFor)  ")
                  : ((kernelInfo[i]->getKernelType() == PARALLEL_REDUCE)
@@ -119,7 +119,7 @@ int main(int argc, char* argv[]) {
              (kernelInfo[i]->getTime() / totalExecuteTime) * 100.0);
     else
       printf("- %s\n%s%c%f%c%" PRIu64 "%c%f%c%f%c%f\n",
-             kernelInfo[i]->getName(),
+             kernelInfo[i]->getName().c_str(),
              (kernelInfo[i]->getKernelType() == PARALLEL_FOR)
                  ? (" (ParFor)  ")
                  : ((kernelInfo[i]->getKernelType() == PARALLEL_REDUCE)
@@ -146,7 +146,7 @@ int main(int argc, char* argv[]) {
     if (kernelInfo[i]->getKernelType() == REGION) continue;
     if (fixed_width)
       printf("- %100s\n%11s%c%15.5f%c%12" PRIu64 "%c%15.5f%c%7.3f%c%7.3f\n",
-             kernelInfo[i]->getName(),
+             kernelInfo[i]->getName().c_str(),
              (kernelInfo[i]->getKernelType() == PARALLEL_FOR)
                  ? (" (ParFor)  ")
                  : ((kernelInfo[i]->getKernelType() == PARALLEL_REDUCE)
@@ -161,7 +161,7 @@ int main(int argc, char* argv[]) {
              (kernelInfo[i]->getTime() / totalExecuteTime) * 100.0);
     else
       printf("- %s\n%s%c%f%c%" PRIu64 "%c%f%c%f%c%f\n",
-             kernelInfo[i]->getName(),
+             kernelInfo[i]->getName().c_str(),
              (kernelInfo[i]->getKernelType() == PARALLEL_FOR)
                  ? (" (ParFor)  ")
                  : ((kernelInfo[i]->getKernelType() == PARALLEL_REDUCE)
--- a/tests/space-time-stack/test_demangling.cpp
+++ b/tests/space-time-stack/test_demangling.cpp
@@ -1,0 +1,80 @@
+#include <iostream>
+#include <regex>
+#include <sstream>
+
+#include "Kokkos_Core.hpp"
+
+#include "utils/demangle.hpp"
+
+struct Tester {
+  struct TagNamed {};
+  struct TagUnnamed {};
+
+  template <typename execution_space>
+  explicit Tester(const execution_space& space) {
+    //! Explicitly launch a kernel with a name and no tag.
+    Kokkos::parallel_for("named kernel",
+                         Kokkos::RangePolicy<execution_space>(space, 0, 1),
+                         *this);
+
+    //! Explicitly launch a kernel with a name and a tag.
+    Kokkos::parallel_for(
+        "named kernel with tag",
+        Kokkos::RangePolicy<execution_space, TagNamed>(space, 0, 1), *this);
+
+    //! Explicitly launch a kernel with no name and no tag.
+    Kokkos::parallel_for(Kokkos::RangePolicy<execution_space>(space, 0, 1),
+                         *this);
+
+    //! Explicitly launch a kernel with no name and a tag.
+    Kokkos::parallel_for(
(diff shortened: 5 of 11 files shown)


romintomasetti:
@vlkale @dalg24 @masterleinad

CMakeLists.txt
```diff
@@ -116,6 +116,9 @@ include_directories(${CMAKE_CURRENT_SOURCE_DIR}/profiling/all)
 set(COMMON_HEADERS_PATH ${CMAKE_CURRENT_BINARY_DIR}/common)
 include_directories(${COMMON_HEADERS_PATH})
 
+# Allow all tools to include any file.
+include_directories(${CMAKE_SOURCE_DIR})
```
dalg24:
```suggestion
include_directories(${CMAKE_CURRENT_SOURCE_DIR})
```

common/utils/demangle.hpp
```diff
@@ -0,0 +1,70 @@
...
+ *       is freed and the returned pointer must be freed by the caller.
+ */
+inline char* demangleName(char* mangledName)
+{
+#if defined(HAVE_GCC_ABI_DEMANGLE)
+    auto demangled_name = demangleNameImpl(mangledName);
+
+    if(demangled_name != nullptr)
+    {
+        std::free(mangledName);
```
dalg24: ?!!

CMakeLists.txt
romintomasetti: Does it really make a difference ? Since its the root `CMakeLists.txt`... Not against the change but I'd like to understand :wink:

common/utils/demangle.hpp
romintomasetti: I know, it's ugly. But it's how it was implemented in `kp_space_time_stack.cpp` before...

I'll have a look to see if I can get rid of it.

CMakeLists.txt
dalg24: It makes your project more "composable", if someone include it in a meta project via add_subdirectory.

common/utils/demangle.hpp
dalg24: This design might fit the current implementation details of space-time-stack, but I don't feel like it is a good idea to elevate it as is for reuse.  I find it too error prone.

romintomasetti: I fully agree!

romintomasetti: Should we refactor a bit `KernelPerformanceInfo` such that it stores the kernel name as a `std::string` instead of a `char*`? It seems it would make this class a bit more readable...

romintomasetti:
```c++
KernelPerformanceInfo(std::string kName, KernelExecutionType kernelType)
```
is the constructor. So it goes from a `std::string` to a `char*` for some reason (that is maybe outdated).

romintomasetti: @dalg24 I've mae `kernelName` a `std::string`, so I could delete the "ugly" `demangleName` :wink:

common/utils/demangle.hpp
```diff
@@ -0,0 +1,51 @@
...
+#if defined(HAVE_GCC_ABI_DEMANGLE)
+    auto demangled_name = demangleNameImpl(mangledName.c_str());
+
+    if(demangled_name != nullptr)
+    {
+        return std::string(demangled_name.get());
+    }
+#endif
+  return mangledName;
+}
```
masterleinad: Is there any good reason not to combine these two functions given that they are very short and handling of the failure needs to be treated in two places this way?

```suggestion
inline std::string demangleName(const std::string& mangledName)
{
#if defined(HAVE_GCC_ABI_DEMANGLE)
   int status = 0;

    std::unique_ptr<char, decltype(&std::free)> demangled_name(
      abi::__cxa_demangle(mangled_name.c_str(), nullptr, nullptr, &status),
      &std::free
    );

    if(status == 0) return std::string(demangled_name.get());
    else demangled_name.reset();
#endif
  return mangledName;
}
```

profiling/simple-kernel-timer/kp_json_writer.cpp
```diff
@@ -57,20 +57,6 @@ inline void write_json(std::ostream& os, KernelPerformanceInfo const& kp,
...
-	for(unsigned int i = 0; i < kernels.size(); i++) {
-		KernelPerformanceInfo* nextKernel = kernels[i];
-
-		if(strcmp(nextKernel->getName(), kernelName) == 0) {
-			return i;
-		}
-	}
-
-	return -1;
-}
```
masterleinad: Please do this cleanup in a separate pull request or at least a separate commit.

masterleinad (commented):
Please make sure that the pull request title and description reflect all the changes you make here.

common/utils/demangle.hpp
romintomasetti: OK, better indeed :heavy_check_mark:

profiling/simple-kernel-timer/kp_json_writer.cpp
romintomasetti: I'll do it in a separate PR :wink:

romintomasetti: See:
- #227

common/utils/demangle.hpp
```diff
@@ -0,0 +1,37 @@
...
+#endif
+
+#if defined(HAVE_GCC_ABI_DEMANGLE)
+#include <cxxabi.h>
+#endif  // HAVE_GCC_ABI_DEMANGLE
+
+namespace KokkosTools
+{
+
+//! Demangme @p mangled_name.
```
romintomasetti:
```suggestion
//! Demangle @p mangled_name.
```

masterleinad:
You have to fix the indentation as well.

common/utils/demangle.hpp
romintomasetti: OK, done :ballot_box_with_check:

common/utils/demangle.hpp
masterleinad, line 1: This file is missing a
```C++
//@HEADER
// ************************************************************************
//
//                        Kokkos v. 4.0
//       Copyright (2022) National Technology & Engineering
//               Solutions of Sandia, LLC (NTESS).
//
// Under the terms of Contract DE-NA0003525 with NTESS,
// the U.S. Government retains certain rights in this software.
//
// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
// See https://kokkos.org/LICENSE for license information.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
//
//@HEADER
```
header.

romintomasetti: Indeed, and done :+1:

masterleinad (approved):
Fine with me.

CMakeLists.txt
```diff
@@ -116,6 +116,9 @@ include_directories(${CMAKE_CURRENT_SOURCE_DIR}/profiling/all)
 set(COMMON_HEADERS_PATH ${CMAKE_CURRENT_BINARY_DIR}/common)
 include_directories(${COMMON_HEADERS_PATH})
 
+# Allow all tools to include any file.
+include_directories(${CMAKE_CURRENT_SOURCE_DIR})
```
dalg24: In isolation this looks reasonable but it clashes with the other include directories (I find it weird that we have to specify `common/` for some and can omit it for the configured files.)

profiling/simple-kernel-timer/kp_reader.cpp
```diff
@@ -64,8 +64,9 @@ int main(int argc, char* argv[]) {
       KernelPerformanceInfo* new_kernel =
           new KernelPerformanceInfo("", PARALLEL_FOR);
       if (new_kernel->readFromFile(the_file)) {
-        if (strlen(new_kernel->getName()) > 0) {
-          int kernelIndex = find_index(kernelInfo, new_kernel->getName());
+        if (!new_kernel->getName().empty()) {
+          int kernelIndex =
+              find_index(kernelInfo, new_kernel->getName().c_str());
```
dalg24: This is silly, please update find_index and take the trailing argument as a string const ref.

romintomasetti: Indeed, and done.

By the way, I think `find_index` is useless if the variable `kernelInfo`, which is a `std::vector`, is changed to being a `std::set` whose comparator compares the time (see what `compareKernelPerformanceInfo` does).

I think the code would be much easier to read.

If you agree I can make the change.

CMakeLists.txt
romintomasetti: What do you propose then?

common/utils/demangle.hpp
```diff
@@ -0,0 +1,49 @@
...
+inline std::string demangleName(const std::string_view mangled_name) {
+#if defined(HAVE_GCC_ABI_DEMANGLE)
+  int status = 0;
+
+  std::unique_ptr<char, decltype(&std::free)> demangled_name(
+      abi::__cxa_demangle(mangled_name.data(), nullptr, nullptr, &status),
+      &std::free);
+
+  if (status == 0) return std::string(demangled_name.get());
+#endif
```
dalg24, line 45: Why bother with a `unique_ptr`?
```diff
diff --git a/common/utils/demangle.hpp b/common/utils/demangle.hpp
index 47cdfeb..b0ce5d9 100644
--- a/common/utils/demangle.hpp
+++ b/common/utils/demangle.hpp
@@ -17,7 +17,6 @@
 #ifndef KOKKOSTOOLS_COMMON_UTILS_DEMANGLE_HPP
 #define KOKKOSTOOLS_COMMON_UTILS_DEMANGLE_HPP

-#include <memory>
 #include <string>

 #if defined(__GXX_ABI_VERSION)
@@ -33,13 +32,14 @@ namespace KokkosTools {
 //! Demangle @p mangled_name.
 inline std::string demangleName(const std::string_view mangled_name) {
 #if defined(HAVE_GCC_ABI_DEMANGLE)
-  int status = 0;
-
-  std::unique_ptr<char, decltype(&std::free)> demangled_name(
-      abi::__cxa_demangle(mangled_name.data(), nullptr, nullptr, &status),
-      &std::free);
-
-  if (status == 0) return std::string(demangled_name.get());
+  int status;
+  char* demangled_name =
+      abi::__cxa_demangle(mangled_name.data(), nullptr, nullptr, &status);
+  if (demangled_name) {
+    std::string ret(demangled_name);
+    std::free(demangled_name);
+    return ret;
+  }
 #endif
   return std::string(mangled_name);
 }

**Review by dalg24 — COMMENTED** [review]
Please post a before/after tool output snippet in the description.

File: common/utils/demangle.hpp (hunk 8 above)
Comment (romintomasetti) [reply common/utils/demangle.hpp:45]: It wasn't crystal-clear to me from the documentation https://gcc.gnu.org/onlinedocs/libstdc++/libstdc++-html-USERS-4.3/a01696.html if the caller would need to free the memory allocated by `abi::__cxa_demangle` if the status is not equal to zero:
> Returns:
    A pointer to the start of the NUL-terminated demangled name, or NULL if the demangling fails. The caller is responsible for deallocating this memory using free.

I just checked with *valgrind* that for a name that cannot be demangled, there is no leak.

File: CMakeLists.txt (hunk 6 above)
Comment (romintomasetti) [reply CMakeLists.txt]: @dalg24 ?

File: profiling/simple-kernel-timer/kp_reader.cpp (hunk 7 above)
Comment (romintomasetti) [reply profiling/simple-kernel-timer/kp_reader.cpp]: @dalg24 I guess you'll want to make the change in another PR :wink:

File: common/utils/demangle.hpp (hunk 8 above)
Comment (romintomasetti) [reply common/utils/demangle.hpp:45]: So I applied your suggestion :wink:

**Comment by romintomasetti** [comment]
@dalg24 @masterleinad 

I added a test for the *space time stack* tool that checks how kernels appear in the report.

So I had to modify slightly the code such that *Kokkos Tools* can also parse *unnamed kernels launched with a work tag*. See the changes for more details.

File: CMakeLists.txt (hunk 6 above)
Comment (dalg24) [reply CMakeLists.txt]: I suppose this is more "contained"
```suggestion
# Allow all tools to include any file.
include_directories(${CMAKE_CURRENT_SOURCE_DIR}/common)
```

common/utils/demangle.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+#if defined(HAVE_GCC_ABI_DEMANGLE)
+#include <cxxabi.h>
+#endif  // HAVE_GCC_ABI_DEMANGLE
+
+namespace KokkosTools {
+
+//! Demangle @p mangled_name.
+inline std::string demangleName(const std::string_view& mangled_name) {
```
dalg24: Why did you change this to pass the `string_view` by const reference?

CMakeLists.txt
romintomasetti: OK, done.

common/utils/demangle.hpp
romintomasetti: Indeed, that doesn't really help. Reverted.

romintomasetti:
@dalg24 @masterleinad Could you please review this one one last time ? :mag: Thanks!

CMakeLists.txt
```diff
@@ -45,6 +45,8 @@ option(KokkosTools_ENABLE_MPI     "Enable MPI support"              OFF)
 option(KokkosTools_ENABLE_CALIPER "Enable building Caliper library" OFF)
 option(KokkosTools_ENABLE_APEX    "Enable building Apex library"    OFF)
 option(KokkosTools_ENABLE_EXAMPLES "Build examples"                 OFF)
+option(KokkosTools_ENABLE_TESTS    "Enable tests"                   OFF)
```
dalg24:
```suggestion
option(KokkosTools_ENABLE_TESTS    "Build tests"                   OFF)
```

tests/space-time-stack/test_demangling.cpp
```diff
@@ -0,0 +1,90 @@
+#include <iostream>
+#include <regex>
+#include <sstream>
+
+#include "Kokkos_Core.hpp"
+#include "kp_all.hpp"
```
dalg24: What do you use from that header?

tests/space-time-stack/test_demangling.cpp
```diff
@@ -0,0 +1,90 @@
...
+                         *this);
+
+    //! Explicitly launch a kernel with no name and a tag.
+    Kokkos::parallel_for(
+        Kokkos::RangePolicy<execution_space, TagUnnamed>(space, 0, size),
+        *this);
+  }
+
+  KOKKOS_INLINE_FUNCTION
+  void operator()(const int index) const {}
```
dalg24:
```suggestion
  KOKKOS_FUNCTION void operator()(int) const {}
```

tests/space-time-stack/test_demangling.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+  template <typename TagType>
+  KOKKOS_INLINE_FUNCTION void operator()(const TagType, const int index) const {
+  }
```
dalg24:
```suggestion
  template <typename Tag>
  KOKKOS_FUNCTION void operator()(Tag, int) const {}
```

tests/space-time-stack/test_demangling.cpp
```diff
@@ -0,0 +1,90 @@
+#include <iostream>
+#include <regex>
+#include <sstream>
+
+#include "Kokkos_Core.hpp"
+#include "kp_all.hpp"
+
+#include "utils/demangle.hpp"
+
+template <typename execution_space, size_t size = 5>
```
dalg24: Why would you want to template the whole struct?
Get rid of `apply()`, move the test to the constructor that you template on `ExecutionSpace`.  Also there is no point on making the number of iteration a template parameter.

dalg24 (commented):
Although obviously adding tests in an improvement but you had almost reached consensus and expanding the scope of your PR will delay merging.

vlkale:
@romintomasetti Sorry to not get back sooner - I was away and on vacation + holiday since this time until late last week, and I am still catching up. The idea of demangle.hpp from your initial post and the current code and CMake in your PR makes sense and looks good to me. My top-level comment is whether you want to test this demangling functionality  for a couple of other tools, e.g., kernel-logger, simple-kernel-timer? You put this in the common subdirectory, so am I correct that it can be used for those too? 

Edit: I see that you have used simple-kernel-timer, so that's great!

Right now, I think
@dalg24 (and @masterleinad) have a better handle on this than me. The PR looks good to me overall. I am good with approving merging this if they are, in order to move things forward.

CMakeLists.txt
romintomasetti: OK, done :heavy_check_mark:

tests/space-time-stack/test_demangling.cpp
romintomasetti: I think it's a leftover. I removed `kp_all.hpp` :wastebasket:

tests/space-time-stack/test_demangling.cpp
romintomasetti: OK. It's indeed implicitly inline since it is defined in the *struct* itself. Done :heavy_check_mark:

romintomasetti: But why would you remove the *const* ?

tests/space-time-stack/test_demangling.cpp
romintomasetti: Done, except for the *const*. Why would you remove it ?

tests/space-time-stack/test_demangling.cpp
romintomasetti: * I think I templated the *struct* over the *execution_space* type to check what the tool would output when looking at kernels defined in templated classes. If you think it is unnecessary, I can get rid of the *execution_space* template.
* For the number of iterations, I think I wanted to have a compile-time range, but I guess it is useless for this test. I'll use a range from 0 to 1 now :wink:

tests/space-time-stack/test_demangling.cpp
dalg24: Because it is just noise.

tests/space-time-stack/test_demangling.cpp
dalg24: Because the const does nothing.  You don't use the argument.

tests/space-time-stack/test_demangling.cpp
```diff
@@ -0,0 +1,80 @@
...
+  template <typename execution_space>
+  explicit Tester(const execution_space& space) {
+    //! Explicitly launch a kernel with a name and no tag.
+    Kokkos::parallel_for("named kernel",
+                         Kokkos::RangePolicy<execution_space>(space, 0, 1),
+                         *this);
+
+    //! Explicitly launch a kernel with a name and a tag.
+    Kokkos::parallel_for(
+        "named kernel",
```
masterleinad:
```suggestion
        "named kernel with tag",
```

masterleinad (commented):
I would prefer to give the functor with a tag a different name. Otherwise, this looks good to me.

tests/space-time-stack/test_demangling.cpp
romintomasetti: OK, done :heavy_check_mark:

masterleinad (approved):

dalg24 (approved):

PR #226, opened 2023-12-27
Don't disable clang-format in kp_json_writer.cpp

masterleinad:
I don't see any good reason to disable clang-format in this file and enforcing the common formatting improves readability.


dalg24 (approved):
I had done that when I introduced that new source file to keep the diff minimal with `kp_reader.cpp`.  That was before we formatted all source files.

Issue #234, opened 2024-01-23
Common `enum FrameType` for consistent "kernel" type treatment


romintomasetti:
As discussed in #218, it would be better if the frame type is an `enum` shared by all tools. See `common/FrameType.hpp` in #218.

> I created a new enum FrameType that seeks to homogenize how kernel types are treated amongst tools, and how they are printed. This also deserves another PR if you agree on the principle.

@crtrott @dalg24 @vlkale If you agree on the principle (and the implementation given in #218) I can make a PR for this.

PR #235, opened 2024-01-30
merge kp-kernels-timers

blegouix:
Environment flag "KOKKOS_TOOLS_TIMER_JSON" determines if export is done in .dat or .json format.

Regions are also exported in the .json (it was not the case previously)


masterleinad:
Can you please describe the behavior when setting `KP_KERNEL_TIMER_JSON` and when you are not setting it? Is something exported? What is printed? How did the behavior change in both cases?

profiling/simple-kernel-timer/kp_kernel_timer.cpp
```diff
@@ -50,25 +55,94 @@ void kokkosp_init_library(const int loadSeq, const uint64_t interfaceVer,
...
+  const char* kp_kernel_timer_json_raw = getenv("KP_KERNEL_TIMER_JSON");
+  const bool kp_kernel_timer_json =
+      strcmp(kp_kernel_timer_json_raw, "1") == 0 ||
+      strcmp(kp_kernel_timer_json_raw, "true") == 0 ||
+      strcmp(kp_kernel_timer_json_raw, "True") == 0;
+
+  double finishTime  = seconds();
+  double kernelTimes = 0;
+
+  char* mpi_rank = getenv("OMPI_COMM_WORLD_RANK");
```
dalg24: I would drop the MPI support from this PR.
This can be handled in a separate PR as needed.

profiling/simple-kernel-timer/kp_kernel_timer.cpp
```diff
@@ -50,25 +55,94 @@ void kokkosp_init_library(const int loadSeq, const uint64_t interfaceVer,
...
 
 void kokkosp_finalize_library() {
-  double finishTime = seconds();
+  const char* kp_kernel_timer_json_raw = getenv("KP_KERNEL_TIMER_JSON");
+  const bool kp_kernel_timer_json =
+      strcmp(kp_kernel_timer_json_raw, "1") == 0 ||
+      strcmp(kp_kernel_timer_json_raw, "true") == 0 ||
+      strcmp(kp_kernel_timer_json_raw, "True") == 0;
+
+  double finishTime  = seconds();
```
dalg24: No.  `finishTime` needs to be recorded first thing.
We do not want to time parsing the environment.

blegouix:
`KP_KERNEL_TIMER_JSON=1, True` or `true` leads to the behavior of previous `kp_kernel_timer_json.cpp` file (export of a .json file giving the timings), `KP_KERNEL_TIMER_JSON` with any other value **or unset** gives the behavior of previous `kp_kernel_timer.cpp` (export of a `.dat` file, readable with `kp_reader`)

Moreover, `regions` timings was not implemented in `kp_kernel_timer_json.cpp`. Now they are.

dalg24 (commented):
avoid duplicating code and explore sharing the implementation with JSON writer.

vlkale:
Thanks @blegouix 

I would like to request changing the environment variable name to `'KOKKOS_TOOLS_TIMER_JSON` so as to be uniform with the other Kokkos Tools environment variables. 

The format for tools environment variables is `KOKKOS_TOOLS_<nameofTool>_<Variable>`. It looks like this is not clear in the documentation of Kokkos Tools, and I will update this from my end.

vlkale:
@blegouix 

I just want to point it out that there seems to be an error in one of the CI build tests. It comes from PAPI connector. The error is in the screenshot below. The CI simple-build (without Kokkos) passes. 

Screenshot 2024-02-01 at 9 24 21 AM

blegouix:
I think the PAPI thing is just a warning, but there is indeed an error there:

`/usr/bin/ld: ../profiling/all/libkokkostools.so: undefined reference to `KokkosTools::KernelTimerJSON::get_event_set()'`

I'll check it, it is possible that namespace changed or something

vlkale:
Great, thanks!

Yeah, the Kokkos Tools CI tests don’t pass unless all warnings are eliminated.

From: Baptiste Legouix ***@***.***>
Date: Thursday, February 1, 2024 at 9:49 AM
To: kokkos/kokkos-tools ***@***.***>
Cc: Vivek Kale ***@***.***>, Comment ***@***.***>
Subject: Re: [kokkos/kokkos-tools] merge kp-kernels-timers (PR #235)

I think the PAPI thing is just a warning, but there is indeed an error there:

/usr/bin/ld: ../profiling/all/libkokkostools.so: undefined reference to KokkosTools::KernelTimerJSON::get_event_set()'`

I'll check it, it is possible that namespace changed or something

vlkale:
> I think the PAPI thing is just a warning, but there is indeed an error there:
> 
> `/usr/bin/ld: ../profiling/all/libkokkostools.so: undefined reference to `KokkosTools::KernelTimerJSON::get_event_set()'`
> 
> I'll check it, it is possible that namespace changed or something

Great, thanks!

Yeah, the Kokkos Tools CI tests don’t pass unless all warnings are eliminated.

blegouix:
Can you run the CI again ?

blegouix:
Can you run the CI again ? Also, I do not think I have the right version of clang-format, could you format it for me ?

Thx,

vlkale:
> Can you run the CI again ? 

Done. 

> Also, I do not think I have the right version of clang-format, could you format it for me ?
> 
> Thx,

Use clang-format-8. 

On a Mac (if you have it), you can use homebrew and do `brew install clang-format@8`. Otherwise, can you download it from there on clang.llvm.org? I think it will help you to get clang-format-8 for PR contribution to Kokkos Tools. If you can't get this to work easily, I can do this for you.

profiling/simple-kernel-timer/kp_kernel_timer.cpp
```diff
@@ -52,23 +57,88 @@ void kokkosp_init_library(const int loadSeq, const uint64_t interfaceVer,
...
+      print_comma = true;
+    }
+
+    fprintf(output_data, "\n");
+    fprintf(output_data, "    ]\n");
 
-  for (auto kernel_itr = count_map.begin(); kernel_itr != count_map.end();
-       kernel_itr++) {
-    kernel_itr->second->writeToBinaryFile(output_data);
+    fprintf(output_data, "}\n}");
```
romintomasetti, line 143: It would be nice to have a test for this. Otherwise, how could a reviewer be sure that nothing broke ?

blegouix:
I have no idea why the remaining test fails :(

profiling/simple-kernel-timer/kp_kernel_timer.cpp
```diff
@@ -52,23 +57,88 @@ void kokkosp_init_library(const int loadSeq, const uint64_t interfaceVer,
 void kokkosp_finalize_library() {
   double finishTime = seconds();
 
+  const char* kokkos_tools_timer_json_raw = getenv("KOKKOS_TOOLS_TIMER_JSON");
+  const bool kokkos_tools_timer_json =
+      strcmp(kokkos_tools_timer_json_raw, "1") == 0 ||
```
dalg24: This will yield a segfault if the environment variable is not defined.

profiling/simple-kernel-timer/kp_kernel_timer.cpp
```diff
@@ -52,23 +57,88 @@ void kokkosp_init_library(const int loadSeq, const uint64_t interfaceVer,
...
+      strcmp(kokkos_tools_timer_json_raw, "true") == 0 ||
+      strcmp(kokkos_tools_timer_json_raw, "True") == 0;
```
blegouix, line 64:
```suggestion
  const bool kokkos_tools_timer_json =
      kokkos_tools_timer_json_raw == NULL ||
      strcmp(kokkos_tools_timer_json_raw, "1") == 0 ||
      strcmp(kokkos_tools_timer_json_raw, "true") == 0 ||
      strcmp(kokkos_tools_timer_json_raw, "True") == 0;
```

profiling/simple-kernel-timer/kp_kernel_timer.cpp
```diff
@@ -52,23 +57,90 @@ void kokkosp_init_library(const int loadSeq, const uint64_t interfaceVer,
...
+      kernel_itr->second->writeToBinaryFile(output_data);
+    }
+  } else {
+    for (auto kernel_itr = count_map.begin(); kernel_itr != count_map.end();
+         kernel_itr++) {
+      kernelList.push_back(kernel_itr->second);
+      kernelTimes += kernel_itr->second->getTime();
+    }
+
+    std::sort(kernelList.begin(), kernelList.end(),
```
romintomasetti, line 97: It seems that sorting the kernels wasn't a feature of `kp_kernel_timer`, but was used in `kp_kernel_timer_json`.

I think sorting or not needs to be clearly stated. What do you think @vlkale?

IMO sorting isn't a requirement and can be done by the user if need be.

profiling/simple-kernel-timer/kp_kernel_timer.cpp
```diff
@@ -52,23 +57,90 @@ void kokkosp_init_library(const int loadSeq, const uint64_t interfaceVer,
...
 
   char* fileOutput = (char*)malloc(sizeof(char) * 256);
-  snprintf(fileOutput, 256, "%s-%d.dat", hostname, (int)getpid());
+  snprintf(fileOutput, 256, "%s-%d.%s", hostname, (int)getpid(),
+           kokkos_tools_timer_json ? "json" : "dat");
 
   free(hostname);
   FILE* output_data = fopen(fileOutput, "wb");
 
+  std::vector<KernelPerformanceInfo*> kernelList;
```
romintomasetti: Seems like the scope of `kernelList` can be reduced.

profiling/simple-kernel-timer/kp_kernel_timer.cpp
```diff
@@ -52,23 +57,90 @@ void kokkosp_init_library(const int loadSeq, const uint64_t interfaceVer,
...
+
+#define KERNEL_INFO_INDENT "       "
+
+    bool print_comma = false;
+    for (auto const& kernel : count_map) {
+      if (!is_region(*std::get<1>(kernel))) continue;
+      if (print_comma) fprintf(output_data, ",\n");
+      kernel.second->writeToJSONFile(output_data, KERNEL_INFO_INDENT);
+      print_comma = true;
+    }
```
romintomasetti, line 125: I would not split the results in "regions" and "non-regions".

As discussed in #234, they are all a "frame". IMO no need to make the JSON file harder to parse, let's keep "regions" and "non-regions" together in *e.g.* a `frame-perf-info` field in the JSON output.

profiling/simple-kernel-timer/kp_kernel_timer.cpp
romintomasetti: I mean, for this tool, sorting by (I guess) increasing time, isn't needed.

profiling/simple-kernel-timer/kp_kernel_timer.cpp
blegouix: Ok, but I would prefer someone else to make this improvement, in this MR or in another one. Actually all I needed was a tool to export regions timer in a .json file easily, so I reproduced the behavior already implemented for the .dat export.

profiling/simple-kernel-timer/kp_kernel_timer.cpp
dalg24: ~~Whether to sort or not is beyond the scope of this PR.~~  I meant to say stick with whatever was done before and we can change if necessary in a follow up

profiling/simple-kernel-timer/kp_kernel_timer.cpp
dalg24: Beyond the scope of this PR.

romintomasetti: Sorry @blegouix @dalg24 !

I thought that `kp_kernel_timer_json` was already catching the regions. But in fact it seems this PR actually *moves* the JSON output logic to `kp_kernel_timer` such that when using the JSON output format it also captures regions.

Sorry for the confusion.

romintomasetti: > I would not split the results in "regions" and "non-regions".
> 
> As discussed in #234, they are all a "frame". IMO no need to make the JSON file harder to parse, let's keep "regions" and "non-regions" together in _e.g._ a `frame-perf-info` field in the JSON output.

Sorry @blegouix @dalg24 !

I thought that `kp_kernel_timer_json` was already catching the regions. But in fact it seems this PR actually moves the JSON output logic to `kp_kernel_timer` such that when using the JSON output format it also captures regions (since `kp_kernel_timer` had always captured regions).

Sorry for the confusion.

IMO the PR description could be updated a bit in light of what I said just above (so as to avoid any misunderstanding of the goal of this PR).

profiling/simple-kernel-timer/kp_kernel_timer.cpp
blegouix: `IMO sorting isn't a requirement and can be done by the user if need be.` just to say, If we can avoid post-processing of those .json and get the information easily just by opening them, this is better. I guess it could be an option KOKKOS_TOOLS_TIMER_SORT though

romintomasetti: IMO if the tool documentation is not saying "results are always sorted by increasing time", it shouldn't be.

Also, maybe you might want to see your regions sorted by increasing time, but I might want to see the regions in their execution order.

So indeed, it seems your `KOKKOS_TOOLS_TIMER_SORT` option would be a good compromise :wink:

blegouix:
Hello, can we merge this or do you think there are remaining things to do ?

vlkale:
Started CI workflow.

I am OK with not sorting the times. I see a broader motivation though - this is how CUDA nvtx tools display timing output, and so there is an expectation by a Kokkos user for this may be similar. 

I think the environment variable mentioned in your last message is a good temporary solution. 

Thanks a lot for your help and work here!

blegouix:
The environment variable is not in the scope of this MR right ?

Im on holidays next week so would be great if we can merge before that

profiling/simple-kernel-timer/kp_kernel_timer.cpp
```diff
@@ -52,23 +57,90 @@ void kokkosp_init_library(const int loadSeq, const uint64_t interfaceVer,
...
+    fprintf(output_data, "{\n\"kokkos-kernel-data\" : {\n");
+    fprintf(output_data, "    \"total-app-time\"         : %10.3f,\n",
+            totalExecuteTime);
+    fprintf(output_data, "    \"total-kernel-times\"     : %10.3f,\n",
+            kernelTimes);
+    fprintf(output_data, "    \"total-non-kernel-times\" : %10.3f,\n",
+            (totalExecuteTime - kernelTimes));
+
+    const double percentKokkos = (kernelTimes / totalExecuteTime) * 100.0;
+    fprintf(output_data, "    \"percent-in-kernels\"     : %6.2f,\n",
```
romintomasetti, line 109: I'm not sure that aggregates should be part of the JSON report.

It makes sense that "console" reports do aggregate times so the user is presented with a nice summary.

But IMO JSON output should be as raw as possible, since I guess the sole purpose of JSON format is to post-process results somewhere else in a user-specific way.

I think it is not valuable that Kokkos Tools adds aggregated metrics in the JSON output...

blegouix: Maybe but it is not in the scope of the MR. It was already there in the timer_json.cpp

(but actually as a user I would say that if you remove aggregated metrics in the output you should provide an easy way to post-process automatically (like execute python script when kokkos finalizes), otherwise it makes things more complicated than they should  for people who just want to monitor the performance of their code in development stage)

dalg24 (approved):
This is good enough for now.

As a follow up I want us to consider using an environment variable (and command-line argument) to select the output format instead of the `KOKKOS_TOOLS_TIMER_JSON` boolean option.
I also want us to reuse the JSON writing code from `kp_json_writer.cpp` instead of duplicating.
And finally we need to add tests for that tool.

PR #237, opened 2024-02-06
tests: add Google Test and use it to test `space-time-stack` connector

romintomasetti:
This PR adds Google Test.

Google Test is then used in `test_demangling.cpp` with the `space-time-stack` tool.

### DIFF
--- a/cmake/BuildGTest.cmake
+++ b/cmake/BuildGTest.cmake
@@ -1,0 +1,35 @@
+# Look for Google Test and enable it as a target.
+#
+# The main targets that will be available are:
+#       * GTest::gtest
+#       * GTest::gmock
+#
+# References:
+#   * https://github.com/google/googletest
+#   * https://matgomes.com/integrate-google-test-into-cmake/
+#   * https://google.github.io/googletest/quickstart-cmake.html
+#   * https://jeremimucha.com/2021/04/cmake-fetchcontent/
+
+include(FetchContent)
+
+# Declare the Google Test dependency
+FetchContent_Declare(
+    googletest
+    GIT_REPOSITORY https://github.com/google/googletest.git
+    GIT_TAG        v1.14.0
+)
+
+# If not yet populated, add Google Test to the build with the following options:
+#       * disable installation of Google Test
+#       * enable GMock
+# Note that we could have used FetchContent_MakeAvailable instead, but it would then
+# use the default configuration that would install Google Test.
+FetchContent_GetProperties(googletest)
+if (NOT googletest_POPULATED)
+    FetchContent_Populate(googletest)
+
+    set(BUILD_GMOCK   ON)
+    set(INSTALL_GTEST OFF)
+
+    add_subdirectory(${googletest_SOURCE_DIR} ${googletest_BINARY_DIR} EXCLUDE_FROM_ALL)
+endif()
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,1 +1,61 @@
+# Add an executable and its related test.
+#
+# The executable is always linked to 'kokkostools' and 'test_common'.
+#
+# Arguments:
+#   TARGET_NAME       : name of the test (required)
+#   SOURCE_FILE       : source file, defaults to <TARGET_NAME>.cpp (optional)
+#   KOKKOS_TOOLS_LIBS : the test environment will received the variable 'KOKKOS_TOOLS_LIBS' that is set as the path
+#                       to the target file of this argument (optional)
+function(kp_add_executable_and_test)
+
+    cmake_parse_arguments(kaeat_args "" "TARGET_NAME;SOURCE_FILE;KOKKOS_TOOLS_LIBS" "" ${ARGN})
+
+    if(NOT DEFINED kaeat_args_TARGET_NAME)
+        message(FATAL_ERROR "'TARGET_NAME' is a required argument.")
+    endif()
+
+    if(NOT DEFINED kaeat_args_SOURCE_FILE)
+        set(kaeat_args_SOURCE_FILE "${kaeat_args_TARGET_NAME}.cpp")
+    endif()
+
+    add_executable(${kaeat_args_TARGET_NAME})
+
+    target_sources(
+        ${kaeat_args_TARGET_NAME}
+        PRIVATE
+            ${kaeat_args_SOURCE_FILE}
+    )
+    target_link_libraries(
+        ${kaeat_args_TARGET_NAME}
+        PRIVATE
+            kokkostools test_common
+    )
+
+    add_test(
+        NAME ${kaeat_args_TARGET_NAME}
+        COMMAND $<TARGET_FILE:${kaeat_args_TARGET_NAME}>
+    )
+
+    if(DEFINED kaeat_args_KOKKOS_TOOLS_LIBS)
+        set_property(
+            TEST ${kaeat_args_TARGET_NAME}
+            APPEND
+            PROPERTY
+                ENVIRONMENT "KOKKOS_TOOLS_LIBS=$<TARGET_FILE:${kaeat_args_KOKKOS_TOOLS_LIBS}>"
+        )
+    endif()
+
+endfunction(kp_add_executable_and_test)
+
+# Create a test library that contains the required Kokkos and Google Test
+# initialization sequence.
+add_library(test_common OBJECT)
+target_sources(
+    test_common
+    PRIVATE
+        UnitTestMain.cpp
+)
+target_link_libraries(test_common PUBLIC GTest::gtest GTest::gmock Kokkos::kokkos)
+
 add_subdirectory(space-time-stack)
--- a/tests/space-time-stack/CMakeLists.txt
+++ b/tests/space-time-stack/CMakeLists.txt
@@ -1,18 +1,5 @@
-# A function to add such "simple" tests in 'tests/CMakeLists.txt' might be a good option.
-add_executable(test_space_time_stack_demangling)
-target_sources(
-    test_space_time_stack_demangling
-    PRIVATE
-        test_demangling.cpp
-)
-target_link_libraries(
-    test_space_time_stack_demangling
-    PRIVATE
-        Kokkos::kokkos kokkostools
-)
-add_test(
-    NAME test_space_time_stack_demangling
-    COMMAND $<TARGET_FILE:test_space_time_stack_demangling>
-        --kokkos-tools-libs=$<TARGET_FILE:kp_space_time_stack>
-        --kokkos-tools-args=1e-9
+kp_add_executable_and_test(
+    TARGET_NAME       test_space_time_stack_demangling
+    SOURCE_FILE       test_demangling.cpp
+    KOKKOS_TOOLS_LIBS kp_space_time_stack
 )
--- a/.github/workflows/build-with-kokkos.yml
+++ b/.github/workflows/build-with-kokkos.yml
@@ -71,9 +71,10 @@ jobs:
                   exit -1
           esac
 
-      - name: Install CMake, OpenMPI, PAPI and dtrace
+      - name: Install git, CMake, OpenMPI, PAPI and dtrace
         run: |
           apt --yes --no-install-recommends install \
+            git ca-certificates \
             cmake make \
             libopenmpi-dev \
             systemtap-sdt-dev \
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -272,6 +272,7 @@ endif()
 # Tests
 if(KokkosTools_ENABLE_TESTS)
   enable_testing()
+  include(cmake/BuildGTest.cmake)
   add_subdirectory(tests)
 endif()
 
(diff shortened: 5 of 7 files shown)


romintomasetti:
@masterleinad Is this version better?

romintomasetti:
@vlkale @dalg24 @masterleinad

vlkale (commented):
@romintomasetti Thanks for this PR on Google Tests! 

My main question/comment on this PR is about whether it serves for testing just the Kokkos Tools connector space time stack, or if some or all this PR can be generalized to work for any Kokkos Tools? It seems like the latter. It would be good to differentiate and factor out tests common across all tools.

romintomasetti:
@vlkale I think it should be used to test all the connectors. As I understand it, it seems the recurring pattern will probably be to have a common "unit test cpp file" that initializes Google Test, but not Kokkos. Because tests will likely check what happens when `Kokkos` finalizes, so each test is likely to initialize and finalize `Kokkos`, as I did in `test_demangling.cpp` in this PR.

I would propose that generalizations come at a later stage, when more connectors are tested. IMO it's not the purpose of this PR to add any additional test :wink:

vlkale:
@romintomasetti I see now, looking again at the most recent comments of PR #236. 

Maybe the title of the PR could say it's solely for space-time-stack (to avoid confusion to others with other future Google Test PRs). Anyway, I see what you're doing here. 

Looking at your files in just this PR and from my understanding of Google Tests, it is good with me.

romintomasetti:
@vlkale Good! I've updated the title :wink:

vlkale:
> @vlkale I think it should be used to test all the connectors. As I understand it, it seems the recurring pattern will probably be to have a common "unit test cpp file" that initializes Google Test, but not Kokkos. Because tests will likely check what happens when `Kokkos` finalizes, so each test is likely to initialize and finalize `Kokkos`, as I did in `test_demangling.cpp` in this PR.
> 
> 
Yes, this is one good approach to the design.

Checking that this indeed makes sense for all other Kokkos Tools connectors (in existence at least) should certainly be a separate PR :)

cmake/BuildGTest.cmake
```diff
@@ -0,0 +1,35 @@
...
+# If not yet populated, add Google Test to the build with the following options:
+#       * disable installation of Google Test
+#       * enable GMock
+# Note that we could have used FetchContent_MakeAvailable instead, but it would then
+# use the default configuration that would install Google Test.
+FetchContent_GetProperties(googletest)
+if (NOT googletest_POPULATED)
+    FetchContent_Populate(googletest)
+
+    set(BUILD_GMOCK   ON)
```
masterleinad, line 31: What do you need `GMOCK` for?

tests/CMakeLists.txt
```diff
@@ -1 +1,50 @@
+# Add one test.
+#
+# The executable is always linked to 'kokkostools' and 'test_common'.
+#
+# Arguments:
+#   TARGET_NAME : name of the test
+#   SOURCE_FILE : source file, defaults to <TARGET_NAME>.cpp
+function(add_one_test)
+
+    cmake_parse_arguments(aot_args "" "TARGET_NAME;SOURCE_FILE" "" ${ARGN})
```
masterleinad: Maybe
```suggestion
    cmake_parse_arguments(TEST_ARGS "" "TARGET_NAME;SOURCE_FILE" "" ${ARGN})
```
to have consistent capitalization?

cmake/BuildGTest.cmake
romintomasetti: In `test_demangling.cpp`:
```c++
#include "gmock/gmock.h"
```
allows the test to look like:
```c++
EXPECT_THAT(output.str(), ::testing::ContainsRegex(matcher));
```

And if I don't enable `GMcok` and link to it:
```bash
/workspaces/kokkos-tools/tests/space-time-stack/test_demangling.cpp:4:10: fatal error: gmock/gmock.h: No such file or directory
    4 | #include "gmock/gmock.h"
```

tests/CMakeLists.txt
romintomasetti: The convention I usually follow for the `PREFIX` given to `cmake_parse_arguments` is the following:
1. Take the first letter of each "word" in the function name (here `*a*dd_*o*ne_*t*est` leads to `aot`).
2. Append `_args`.

I think there are 2 advantages using this convention:
1. You reduce the probability that the chosen prefix matches a variable from the parent scope (note how easy it would be to get someone define `TEST_ARGS` somewhere in the parent scope for instance, while `aot_args` seems less probable IMO).
2. You get a consistent and indisputable recipe for the `PREFIX`.

What do you think?

romintomasetti: IMO having the `PREFIX` in lower case also helps reading the function's code :eyeglasses:

tests/TestBase.cpp
```diff
@@ -0,0 +1,14 @@
+#include "gtest/gtest.h"
+
+#include "Kokkos_Core.hpp"
+
+//! Main entry point for tests.
+int main(int argc, char* argv[]) {
+  //! Initialize Google Test.
+  ::testing::InitGoogleTest(&argc, argv);
+
+  //! Run tests.
```
dalg24: All these comments are kinda pointless, they are stating the obvious.

tests/TestBase.cpp
dalg24, line 1: I think "base" does not necessarily convey that it contains the `main` function.

tests/CMakeLists.txt
```diff
@@ -1 +1,50 @@
+# Add one test.
+#
+# The executable is always linked to 'kokkostools' and 'test_common'.
+#
+# Arguments:
+#   TARGET_NAME : name of the test
+#   SOURCE_FILE : source file, defaults to <TARGET_NAME>.cpp
+function(add_one_test)
```
dalg24: Consider calling it `kp_add_executable_and_test`

tests/space-time-stack/CMakeLists.txt
```diff
@@ -1,18 +1,10 @@
...
-target_sources(
-    test_space_time_stack_demangling
-    PRIVATE
-        test_demangling.cpp
+add_one_test(
+    TARGET_NAME test_space_time_stack_demangling
+    SOURCE_FILE test_demangling
 )
-target_link_libraries(
+target_compile_definitions(
```
dalg24: Why are you using a compile definition for this?
It does not look like the right tool for the job.
Why not use a command line argument or an environment variable?

tests/TestBase.cpp
romintomasetti: I will rename it `UnitTestMain.cpp` as in `Kokkos` :wink:

tests/CMakeLists.txt
romintomasetti: Ok, consider it done.

tests/space-time-stack/CMakeLists.txt
romintomasetti: For now, the test looks like this:
```c++
TEST(SpaceTimeStackTest, demangling) {
  //! Fake arguments a user would give via CLI.
  std::string arg{"--kokkos-tools-libs=" LIBKP_SPACE_TIME_STACK};
  int argc = 1;
  std::vector<char*> argv{arg.data(), nullptr};

  //! Initialize @c Kokkos.
  Kokkos::initialize(argc, argv.data());
```
and so it's true that the path to the library is carved in stone. I would argue that it's no big deal since (my guess is) no one will ever use this test with another library :angel:

But I'm OK with changing the approach to something more flexible :wink:

Are you suggesting that I should not hardcode the library path in the executable, and rather read it from the environment ? Something like what follows:
```c++
TEST(SpaceTimeStackTest, demangling) {
  //! Fake arguments a user would give via CLI.
  std::string arg{"--kokkos-tools-libs=" + getenv("LIBKP_SPACE_TIME_STACK")};
  int argc = 1;
  std::vector<char*> argv{arg.data(), nullptr};

  //! Initialize @c Kokkos.
  Kokkos::initialize(argc, argv.data());
```
with the test having additional environment like:
```cmake
set_property(
        TEST "..."
        APPEND
        PROPERTY
            ENVIRONMENT "LIBKP_SPACE_TIME_STACK=..."
    )
```

dalg24: Kokkos looks for `KOKKOS_TOOLS_LIBS`.  We don't need to invent a new way to enable.

tests/space-time-stack/CMakeLists.txt
```diff
@@ -1,18 +1,10 @@
 # A function to add such "simple" tests in 'tests/CMakeLists.txt' might be a good option.
-add_executable(test_space_time_stack_demangling)
-target_sources(
-    test_space_time_stack_demangling
-    PRIVATE
-        test_demangling.cpp
+kp_add_executable_and_test(
+    TARGET_NAME test_space_time_stack_demangling
+    SOURCE_FILE test_demangling
```
masterleinad: Why doesn't this need to be
```suggestion
    SOURCE_FILE test_demangling.cpp
```
?

tests/CMakeLists.txt
```diff
@@ -1 +1,50 @@
...
+
+    cmake_parse_arguments(kaeat_args "" "TARGET_NAME;SOURCE_FILE" "" ${ARGN})
+
+    if(NOT DEFINED kaeat_args_TARGET_NAME)
+        message(FATAL_ERROR "Wrong usage.")
+    endif()
+
+    if(NOT DEFINED kaeat_args_SOURCE_FILE)
+        set(kaeat_args_SOURCE_FILE "${kaeat_args_NAME}.cpp")
+    endif()
```
masterleinad: This isn't used and it's not clear to me that setting a default source file improves behavior (instead of just erroring out).

tests/space-time-stack/CMakeLists.txt
romintomasetti: I've added `.cpp`. But for posterity, it seems `CMake` will try to find your file by testing several suffixes:
```bash
-- Configuring done (1.1s)
CMake Error at tests/CMakeLists.txt:22 (target_sources):
  Cannot find source file:

    test_demanglig

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .mpp .m .M .mm .ixx .cppm
  .ccm .cxxm .c++m .h .hh .h++ .hm .hpp .hxx .in .txx .f .F .for .f77 .f90
  .f95 .f03 .hip .ispc
Call Stack (most recent call first):
  tests/space-time-stack/CMakeLists.txt:2 (kp_add_executable_and_test)


-- Generating done (0.0s)
```

tests/CMakeLists.txt
romintomasetti: OK, removed.

tests/space-time-stack/CMakeLists.txt
romintomasetti: @dalg24 Indeed!

I've changed `kp_add_executable_and_test` such that it also takes the `CONNECTOR_LIB`. The test environment is enriched with `"KOKKOS_TOOLS_LIBS=$<TARGET_FILE:${kaeat_args_CONNECTOR_LIB}>"`.

Is this approach what you had in mind ?

tests/CMakeLists.txt
```diff
@@ -1 +1,56 @@
...
+    add_test(
+        NAME ${kaeat_args_TARGET_NAME}
+        COMMAND $<TARGET_FILE:${kaeat_args_TARGET_NAME}>
+    )
+    set_property(
+        TEST ${kaeat_args_TARGET_NAME}
+        APPEND
+        PROPERTY
+            ENVIRONMENT "KOKKOS_TOOLS_LIBS=$<TARGET_FILE:${kaeat_args_CONNECTOR_LIB}>"
+    )
```
dalg24: Why is it not guarded with `if(kaeat_args_CONNECTOR_LIB)`?
Is there a good reason to make it required to pass that argument?

tests/CMakeLists.txt
dalg24: FWIW I also dislike the prefix you picked
(Not blocking though)

tests/CMakeLists.txt
```diff
@@ -1 +1,56 @@
...
+#
+# The executable is always linked to 'kokkostools' and 'test_common'.
+#
+# The test environment always contains 'KOKKOS_TOOLS_LIBS' that is set as the path
+# to the target file of 'CONNECTOR_LIB'.
+#
+# Arguments:
+#   TARGET_NAME   : name of the test
+#   SOURCE_FILE   : source file, defaults to <TARGET_NAME>.cpp
+#   CONNECTOR_LIB : connector library target name under test
```
dalg24: Why not just call it `KOKKOS_TOOLS_LIBS` or `TOOLS_LIBS`?
It makes it self documented.
Also "connector" has been used to refer to some tools that inject user-defined annotations into vendor tools.  Clearly here that argument is intended for any tool.

romintomasetti: Indeed, you're right that "connector" is not a good choice.

So I've changed to `KOKKOS_TOOLS_LIBS` as you suggested, and it's now an optional argument :ballot_box_with_check:

tests/CMakeLists.txt
romintomasetti: I think that a lot of (not-yet-written) tests will need this environment variable... But I've made it optional :ballot_box_with_check:

masterleinad (approved):
Looks OK to me.

dalg24 (approved):

Issue #147, opened 2022-10-18
Using tools induces undesirable fences in certain cases


vlkale:
# Description of Bug 

When using profiling tools of Kokkos to profile a Kokkos application, Kokkos invokes memory fences that are unneeded. This is particularly the case when a memory fence in a Kokkos application code is itself being profiled. These unneeded memory fences used by the tool are not desirable because they slow down execution of the application when the Kokkos tools are being used and do not reflect the actual performance of the application code. If this problem is not fixed, then users of profiling tools will not obtain an accurate performance profile of the application's execution. Changing the Kokkos tools libraries by removing, or filtering out, the memory fences under particular context where they are not needed will not impact the correctness of the application code. A solution is to filter out the unneeded fences under particular contexts.

# Details of the Problem 

This problem is applicable to Kokkos v. 3.7.0 on any platform. More information and example code for reproducing this problem will be added soon by Vivek. 

From @crtrott and expanded on by @vlkale, there are actual two orthogonal [sub]issues within this GitHub Issue: 

* _Sub-Issue 1_: Some tools need to turn off fencing of tool callbacks, depending on how the tool is used. These tools (like the space-time-stack and simple kernel timer) need to be able to read an environment variable which sets the ON/OFF behavior for the tool setting callback. Specifically, spaceTimeStack and the vendor tools need that. [IOW, we need to eliminate memory fences so that Kokkos tool callbacks can capture the state of environment variables which control behavior of Kokkos-associated tools, e.g., a simple kernel timer, and other vendor tools.]

* _Sub-Issue 2_: If tool callback fencing is on _and_ the tool implements the callback for fencing, we want to filter out fences which are solely due to callback fencing [as they will slow down execution of the application]. Note these are different things: a callback for "beginParallelFor" invokes a Kokkos global fence if callback fencing is on i.e., calls Kokkos::fence(), but a global fence _also_ invokes the tools callback for "beginFence" and "endFence". [IOW, we need to remove the extraneous memory fences of runtime functions of the Kokkos-expanded  code, which invoked by Kokkos's tool callback fencing mode being on, since the tool is implementing the callback for fencing itself]. 

To illustrate: with tools callback fencing _*on*_, the following Kokkos program:

```
Kokkos::fence();

Kokkos::parallel_for(...);

```

expands conceptually into:

```
// Kokkos::fence(); 
beginFence();  // <--- Note: could have a preFence() to set up data structures and bookkeeping for fence
[...]

PR #242, opened 2024-02-28
Add kp_functor_size: print parallel functor sizes

cwpearson:
Prints the count, size, and name of functors passed to Kokkos parallel regions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -147,6 +147,7 @@ add_subdirectory(debugging/kernel-logger)
 
 # Profilers
 if(NOT WIN32)
+  add_subdirectory(profiling/functor-size)
   add_subdirectory(profiling/simple-kernel-timer)
   add_subdirectory(profiling/memory-hwm)
   if(KokkosTools_ENABLE_MPI)
--- a/README.md
+++ b/README.md
@@ -60,6 +60,10 @@ The following provides an overview of the tools available in the set of Kokkos T
 
     Prints Kokkos Kernel and Region events during runtime.
 
++ [**Functor Size**](https://github.com/kokkos/kokkos-tools/wiki/FunctorSize)
+
+    Prints information about the size of the functor objects passed to Kokkos parallel regions.
+
 ### 3rd Party Profiling Tool Hooks
 + [**VTuneConnector:**](https://github.com/kokkos/kokkos-tools/wiki/VTuneConnector)
     
--- a/profiling/all/impl/Kokkos_Profiling_C_Interface.h
+++ b/profiling/all/impl/Kokkos_Profiling_C_Interface.h
@@ -40,17 +40,11 @@ struct Kokkos_Profiling_SpaceHandle {
   char name[64];
 };
 
-const static int Kokkos_Profiling_Kernel_Static_Info_Size = 512;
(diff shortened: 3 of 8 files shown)


cwpearson:
Corresponding Core PR: https://github.com/kokkos/kokkos/pull/6844

vlkale:
@crtrott  I discussed this with Carl. Carl is using this for his work. This would  also well-serve Kokkos Tools because it is a basis to have a generalized ToolState struct that the Kokkos Tools users can use to obtain profiling information per kernel invocation. This ToolsState came out of a discussion from @jrmadsen in #147.  

I looked over this and it looks like the implementation is correct. The only question is whether we want to merge this, or if we want to provide a generalized tool callback than one marking the functor size - maybe something like `kp_mark_kernel_info()` (the kernel is indexed by the kernel ID passed in as a parameter to the function callback). 

One note to consider (@cwpearson - we didn't discuss this) is that functor size is compile-time information and doesn't change dynamically at runtime. I wonder whether doing this in tool callbacks is the only way. 

This PR forms the basis for dynamically changing information/properties of a Kokkos kernel.

cwpearson:
@vlkale I changed the name of the callback to mark_kernel_static_info and defined a 512-byte struct, so while this currently only tracks functor size, it could be extended to include other info.

vlkale:
> @vlkale I changed the name of the callback to mark_kernel_static_info and defined a 512-byte struct, so while this currently only tracks functor size, it could be extended to include other info.

Great, thanks!

CMakeLists.txt
```diff
@@ -147,6 +147,7 @@ add_subdirectory(debugging/kernel-logger)
 
 # Profilers
 if(NOT WIN32)
+  add_subdirectory(profiling/functor-size)
```
romintomasetti, line 150: Please add a test so that:
1. your feature is proven to work as intended
2. we can better understand how you want to use it

To me, it's still a bit unclear at the moment...

romintomasetti (changes requested):
https://github.com/kokkos/kokkos-tools/pull/242/files#r1539554194

CMakeLists.txt
masterleinad: See https://github.com/kokkos/kokkos/pull/6844#issuecomment-2007571734.

romintomasetti: OK, thanks for the pointer.

Still, I think a test must be added in `Kokkos Tools` itself (even if the feature is considered experimental) :wink:

vlkale: @cwpearson I didn't say this in my review of your PR but yes @romintomasetti is right. We have recently been pushing for testing in other Kokkos Tools connectors. I think having a simple test using Google Test would be good. You can see the test from space time stack in Kokkos Tools `develop` branch by @romintomasetti. I can help you with this if you need.



@romintomasetti I think @cwpearson  has demonstrated the use cases reasonably well. The PR from  @masterleinad, see the table of output that @cwpearson has put in.


Perhaps @cwpearson can check if he needs to update the `example` directory of Kokkos Tools, but I don't think he needs to. The Kokkos application programmer need not make any code changes to use this particular tools capability.

PR #244, opened 2024-03-12
GitHub CI: Don't use rocThrust in HIP builds

masterleinad:
The GitHub CI is failing since we are now expecting to find `rocThrust` for HIP by default in `Kokkos` `develop`. Fix this by not explicitly disabling support for `rocThrust`. Installing `rocThrust` would make the image too large for `GitHub` CI.

### DIFF
--- a/kokkos.presets.json
+++ b/kokkos.presets.json
@@ -31,9 +31,10 @@
             "name"        : "ROCm",
             "inherits"    : "default",
             "cacheVariables" : {
-                "Kokkos_ENABLE_HIP"   : "ON",
-                "Kokkos_ARCH_VEGA906" : "ON",
-                "CMAKE_CXX_COMPILER"  : "hipcc"
+                "Kokkos_ENABLE_HIP"       : "ON",
+                "Kokkos_ENABLE_ROCTHRUST" : "OFF",
+                "Kokkos_ARCH_VEGA906"     : "ON",
+                "CMAKE_CXX_COMPILER"      : "hipcc"
             }
         }
     ],


vlkale (commented):
LGTM, and thanks for that!

I was just now trying to figure out how to  fix. This looks like the right solution. 

I don’t know if it’s obvious that rocThrust - a parallel algorithm library for AMD GPUs - is needed to do build testing of Kokkos Tools. Maybe we document or note it in the file of this commit what it is and that it’s needed. I wonder if there is a way to remove this dependency. It’s possible I am missing something.

masterleinad:
> I don’t know if it’s obvious that rocThrust - a parallel algorithm library for AMD GPUs - is needed to do build testing of Kokkos Tools. Maybe we document or note it in the file of this commit what it is and that it’s needed. I wonder if there is a way to remove this dependency. It’s possible I am missing something.

I updated the pull request. It's easier to just disable `rocThrust` support.

vlkale:
> > I don’t know if it’s obvious that rocThrust - a parallel algorithm library for AMD GPUs - is needed to do build testing of Kokkos Tools. Maybe we document or note it in the file of this commit what it is and that it’s needed. I wonder if there is a way to remove this dependency. It’s possible I am missing something.
> 
> 
> 
> I updated the pull request. It's easier to just disable `rocThrust` support.

Got it. Thanks! Yes, I think just disabling is adequate. I see it is passing the build testing now.

masterleinad:
@vlkale ping Can you approve and merge? We can try building with `rocThrust` later if we deem that useful but it's annoying to have the CI fail.

vlkale (approved):

romintomasetti (approved):

romintomasetti:
I think you need a second approval. @crtrott or @dalg24 ?

PR #247, opened 2024-04-04
Remove tool programming interface pointer in kp_sampler_skip.cpp and kp_core.hpp

vlkale:
The Kokkos Core interface passes the struct by value not by pointer.

### DIFF
--- a/common/kokkos-sampler/kp_sampler_skip.cpp
+++ b/common/kokkos-sampler/kp_sampler_skip.cpp
@@ -71,7 +71,7 @@ void kokkosp_provide_tool_programming_interface(
           "KokkosP: Note: Number of functions in Tools Programming Interface "
           "is 0!\n");
   }
-  tpi_funcs = *funcsFromTPI;
+  tpi_funcs = funcsFromTPI;
 }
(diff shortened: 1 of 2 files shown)


crtrott (approved):

dalg24 (commented):
Please provide a proper description.  Removing the pointer because ...

vlkale:
> Please provide a proper description. Removing the pointer because ...

@dalg24  Thanks. I put this in the description (fence function pointer is null - note this impacts the tools that use the tool-invoked fence which is currently just the sampler).

masterleinad (commented):
Can you add a test that would have detected the problem?

vlkale:
> Can you add a test that would have detected the problem?

@masterleinad Thanks for bringing this up. I _think_ that this is handled by that test for the sampler you helped me. 

That said, I can certainly add in a test case in the small chain of patches I have upcoming right now. 

The next one consists of the tests and I will put a matcher to check for the print of the fence being null.

crtrott:
Modified the comment, we will have the test come in with the sampler thing.

PR #213, opened 2023-10-13
Use probability sampling over periodic sampling

vlkale:
Fix #180 . 

This PR is related to old PR #181 

The current Kokkos sampler utility uses periodic sampling via a sampler skip rate. Doing this is often restrictive when sampling profiling and debugging data. For example, doing this can miss out on important data not in the periodicity of kernel invocations. The goal of this PR for the Kokkos sampler  utility is to allow user to use random sampling primarily and periodic sampling secondarily via environment variables in the form `KOKKOS_TOOLS_SAMPLER_xyz`. 

Since the solution should not allow for a combination of both periodicity and probability, the probability will always be chosen. 

For example, let us say that a user requests a a `Kokkos::parallel_for()` every 20th invocation of that `Kokkos::parallel_for()` and requests gather time spent on `Kokkos::parallel_for()` with probability 63% on each invocation of that `Kokkos::parallel_for()`. Then, the sampler will not _skip *trying* to time_ any invocations of the Kokkos::parallel_for() but it _will_ obtain a timing with probability 63% on each invocation of that `Kokkos::parallel_for()`. 

See the `common/kokkos-sampler/README.md` directory for a high-level overview - in English - of the changes.

For later: I will put in slide in the Kokkos Tools tutorial slide on sampling and filtering to explain how to use these utilities.

PR #248, opened 2024-04-04
Put in tests for sampler 

vlkale:
This PR addresses the broader Github Issue #191, which is important for the open-source development of Kokkos Tools. 

The purpose of this PR is to enhance and ensure the Kokkos Tools sampler robustness by adding tests for it. Testing the sampler is important given that: 

(1) It is a parent tool to other tools connectors, i.e., other tools using it will not work if this tool does not call them correctly;
(2) It uses the parts of Kokkos Tools infrastructure that the sampler must exercise to maintain low overhead, i.e., tool-invoked fencing.

This PR adds a directory in tests for the sampler utility and tests that the sampler works correctly. Specifically, it checks that: 

1. every begin callback from the sampler has a matching end callback; and
2. the sampler chooses the correct kernel IDs to sample based on the skip rate

The tests use CTest, and tries to follow the demangling tests for space time stack, but there are important changes needed to the general CMakeLists.txt file to facilitate testing of the sampler. 

Note that corresponding testing will be needed for the kernel filter, which would be a separate PR.  Testing for all tool connectors and this needs to be a separate or multiple separate PRs.


vlkale:
@masterleinad When you were saying to test tool-invoked fencing function pointer is NULL, I wasn't sure if you were referring to just a test that the sampler finished successfully, or if you wanted to directly make sure the pointer wasn't NULL. 

See a possible patch to this PR: https://github.com/vlkale/kokkos-tools/tree/test-for-sampler-patch-add-toolfence-nullptr-test

I just added this during this afternoon, in case you were requesting this (and your comment makes sense and we should test). 

This PR (and not the patch to it) should handle your comment from #247, given the logic in the sampler code. Of course, we need to also test each child library, but that is separate from this PR.

README.md
```diff
@@ -33,11 +33,14 @@ void foo() {
...
 
 ### Utilities
-
 + [**KernelFilter:**](https://github.com/kokkos/kokkos-tools/wiki/KernelFilter) 
 
     A tool which is used in conjunction with analysis tools, to restrict them to a subset of the application.
 
++ [**KernelSampler:**](https://github.com/kokkos/kokkos-tools/wiki/KernelSampler)
+
+   A tool which is used in conjunction with analysis tools, to restrict the tooling to samples of Kokkos kernel invocations.
```
masterleinad:
```suggestion
   A tool to be used in conjunction with analysis tools to restrict the tooling to samples of Kokkos kernel invocations.
```

debugging/kernel-logger/kp_kernel_logger.cpp
```diff
@@ -100,7 +100,7 @@ extern "C" void kokkosp_begin_parallel_scan(const char* name,
   printf("    %s\n", name);
 }
 
-extern "C" void kokkospk_end_parallel_scan(const uint64_t kID) {
+extern "C" void kokkosp_end_parallel_scan(const uint64_t kID) {
```
masterleinad: Part of a different pull request. Please rebase once that's merged or remove duplicated changes.

tests/CMakeLists.txt
```diff
@@ -5,11 +5,14 @@
 # Arguments:
 #   TARGET_NAME       : name of the test (required)
 #   SOURCE_FILE       : source file, defaults to <TARGET_NAME>.cpp (optional)
-#   KOKKOS_TOOLS_LIBS : the test environment will received the variable 'KOKKOS_TOOLS_LIBS' that is set as the path
+#   KOKKOS_TOOLS_LIBS : the test environment will receive the variable 'KOKKOS_TOOLS_LIBS' that is set as the path
 #                       to the target file of this argument (optional)
-function(kp_add_executable_and_test)
+#   KOKKOS_TOOLS_SAMPLER_VERBOSE : the test environment will receive the variable 'KOKKOS_TOOLS_SAMPLER_VERBOSE' that is set as the value of 1 for printing the sample has been taken 
+#   KOKKOS_TOOLS_GLOBALFENCES : test environment receives the variable 'KOKKOS_TOOLS_GLOBALFENCES' that is set as the value of 1 to turn the tool's auto-fencing on. 
+#   KOKKOS_TOOLS_SAMPLER_SKIP : test environment receives the variable 'KOKKOS_TOOLS_SAMPLER_SKIP' that is set as the value of the number of Kokkos kernel invocations to skip before a tooling activity is invoked.
```
masterleinad, line 12: Is this new? If so, are these changes necessary for this pull request?

tests/CMakeLists.txt
```diff
@@ -5,11 +5,14 @@
...
+#   KOKKOS_TOOLS_LIBS : the test environment will receive the variable 'KOKKOS_TOOLS_LIBS' that is set as the path
 #                       to the target file of this argument (optional)
-function(kp_add_executable_and_test)
+#   KOKKOS_TOOLS_SAMPLER_VERBOSE : the test environment will receive the variable 'KOKKOS_TOOLS_SAMPLER_VERBOSE' that is set as the value of 1 for printing the sample has been taken 
+#   KOKKOS_TOOLS_GLOBALFENCES : test environment receives the variable 'KOKKOS_TOOLS_GLOBALFENCES' that is set as the value of 1 to turn the tool's auto-fencing on. 
+#   KOKKOS_TOOLS_SAMPLER_SKIP : test environment receives the variable 'KOKKOS_TOOLS_SAMPLER_SKIP' that is set as the value of the number of Kokkos kernel invocations to skip before a tooling activity is invoked. 
 
-    cmake_parse_arguments(kaeat_args "" "TARGET_NAME;SOURCE_FILE;KOKKOS_TOOLS_LIBS" "" ${ARGN})
+function(kp_add_executable_and_test)
+    cmake_parse_arguments(kaeat_args "" "TARGET_NAME;SOURCE_FILE;KOKKOS_TOOLS_SAMPLER_VERBOSE;KOKKOS_TOOLS_GLOBALFENCES;KOKKOS_TOOLS_SAMPLER_SKIP" "KOKKOS_TOOLS_LIBS" ${ARGN})
```
masterleinad, line 15: Please explain what changed here.

masterleinad (commented):
Please clearly summarize all the changes you are making in this pull request.

dalg24 (commented):
Avoid duplicating code between tests
Your test is really weak.  It will pass if the sampler does nothing and just always forward to the underlying tool.

debugging/kernel-logger/kp_kernel_logger.cpp
vlkale: OK, will do - I just now see that this changed is merged.

README.md
vlkale: This is fine. I was trying to match the kernel filter description for now, but I would say the kernel filter utility wording needs to be fixed too. I think we can make changes on the other parts of the README as a  another PR.

vlkale:
> Avoid duplicating code between tests 

OK. I will check and refactor. 

> Your test is really weak. It will pass if the sampler does nothing and just always forward to the underlying tool.

@dalg24  Thanks. However, I do not believe this is the case: It will not pass if the sampler does nothing. In my test code, _all_  matchers must be seen for this to pass. Also, the EXPECT_THAT function I have will cause a test to fail if the regular expression is not seen. This is the pertinent Google Tests documentation

https://google.github.io/googletest/reference/matchers.html

tests/CMakeLists.txt
vlkale: These are the environment variables needed for the sampler test. One needs to add these in order to control parameters of the sampler utility. Also, the other demangling test doesn't use global fences. For the sampler's test, we need to use global fences to test the sampler.

vlkale:
> > Avoid duplicating code between tests
> 
> OK. I will check and refactor.
> 
> > Your test is really weak. It will pass if the sampler does nothing and just always forward to the underlying tool.
> 
> @dalg24 Thanks. However, I do not believe this is the case: It will not pass if the sampler does nothing. In my test code, _all_ matchers must be seen for this to pass. Also, the EXPECT_THAT function I have will cause a test to fail if the regular expression is not seen. This is the pertinent Google Tests documentation
> 
> https://google.github.io/googletest/reference/matchers.html

@dalg24 On second thought, you may be right - I re-read in the documentation that the matcher matches a single argument. The safest way to fix this is to just match the exact and full string I want to match in one matcher. 

@masterleinad hoping this sounds good to you since you and I talked about this previously.

masterleinad:
> The safest way to fix this is to just match the exact and full string I want to match in one matcher.

You would not even need a regex. Additionally, you can count the number of occurrences of, e.g., "calling child-begin function" to make sure only that the desired kernels were forwarded.

tests/CMakeLists.txt
vlkale: Yes, this is new (though you and I talked about it if you recall in #213). We need these changes in this PR because the environment variables are used to dictate the skip rate, fencing, and sampling print's needed for running the test case. 

I wrote this in the description above.

vlkale:
> Please clearly summarize all the changes you are making in this pull request.

See the revision above.

vlkale:
@dalg24 @masterleinad I believe this should address your comments now. The CI tests are passing (and those CI tests run the tests for the sampler). I also included the null fence pointer test in a more succinct way, and so we should not need that patch I mentioned anymore.

@dalg24 Regarding duplicated code, we could extract out some of the common code across the three test_parxyz.cpp files for brevity. This is working though and has tests that catch the sampler not working. Perhaps we think about this for another PR. 

It would be good to have this looked over and make sure there is nothing else incorrect or lacking for what it is trying to achieve, as laid out in the description.

vlkale:
> > The safest way to fix this is to just match the exact and full string I want to match in one matcher.
> 
> You would not even need a regex. Additionally, you can count the number of occurrences of, e.g., "calling child-begin function" to make sure only that the desired kernels were forwarded.

I have incorporated your feedback, except for the additionally part. I think that can go in a separate PR. 

SYK, I actually tried to use `::testing::Contains.Times(static_cast<int>(2)`, "calling child-begin function.") in this PR, but it was causing a compiler error due to Times being overloaded function. I had initially not put the cast - doing that, it still doesn't compile. This is in the commit I reverted before this last commit.

dalg24:
> > > Your test is really weak. It will pass if the sampler does nothing and just always forward to the underlying tool.
> > 
> > 
> > @dalg24 Thanks. However, I do not believe this is the case: It will not pass if the sampler does nothing. In my test code, _all_ matchers must be seen for this to pass. Also, the EXPECT_THAT function I have will cause a test to fail if the regular expression is not seen. This is the pertinent Google Tests documentation
> > https://google.github.io/googletest/reference/matchers.html
> 
> @dalg24 On second thought, you may be right - I re-read in the documentation that the matcher matches a single argument. The safest way to fix this is to just match the exact and full string I want to match in one matcher.

Assume we merge your test as it is proposed now and a follow PR includes this change.
```diff
diff --git a/common/kokkos-sampler/kp_sampler_skip.cpp b/common/kokkos-sampler/kp_sampler_skip.cpp
index 78cbbff..86b55e3 100644
--- a/common/kokkos-sampler/kp_sampler_skip.cpp
+++ b/common/kokkos-sampler/kp_sampler_skip.cpp
@@ -193,7 +193,7 @@ void kokkosp_begin_parallel_for(const char* name, const uint32_t devID,
                                 uint64_t* kID) {
   *kID                          = uniqID++;
   static uint64_t invocationNum = 0;
-  ++invocationNum;
+  invocationNum;
   if ((invocationNum % kernelSampleSkip) == 0) {
     if (tool_verbosity > 0) {
       std::cout << "KokkosP: sample " << *kID
```
Would you not want your test to fail?

tests/sampler/test_parscan.cpp
```diff
@@ -0,0 +1,83 @@
...
+struct Tester {
+  template <typename execution_space>
+  explicit Tester(const execution_space& space) {
+    //! Explicitly launch a kernel with a name, and run it 150 times with kernel
+    //! logger. Use a periodic sampling with skip rate 51. This should print
+    //! out 2 invocations, and there is a single matcher with a regular
+    //! expression to check this.
+
+    long int N = 1024;
+    int64_t result;
```
dalg24: Won't compile on some platforms...
```suggestion
    long int result;
```
You must match the functor argument that is passed by reference.

tests/CMakeLists.txt
masterleinad: Why do we need global fences here? In the end, we only care about the strings printed and not if the kernels actually ran, right? It would be different if the kernels themselves were printing stuff but all kernels are empty anyway.

common/kokkos-sampler/kp_sampler_skip.cpp
```diff
@@ -1,323 +1,349 @@
...
+  for (int i = 0; i < loadSeq; i++) {
+    nextLibrary = strtok(NULL, ";");
+  }
+
+  nextLibrary = strtok(NULL, ";");
+
+  if (NULL == nextLibrary) {
+    std::cout << "KokkosP: FATAL: No child library of sampler utility library "
+                 "to call in "
+              << profileLibrary << "!\n";
```
masterleinad, line 118: If this is a fatal error, we should abort (and print to std::cerr) instead, right?

common/kokkos-sampler/kp_sampler_skip.cpp
```diff
@@ -1,323 +1,349 @@
...
+  if (tpi_funcs.fence != nullptr) {
+    tpi_funcs.fence(devID);
+    if (tool_verbosity > 1) {
+      std::cout << "KokkosP: Sampler utility sucessfully invoked tool-induced "
+                   "fence on device "
+                << getDeviceID(devID) << ".\n";
+    }
+  } else {
+    std::cout << "KokkosP: FATAL: Kokkos Tools Programming Interface's "
+                 "tool-invoked Fence is NULL!\n";
```
masterleinad, line 60:
```suggestion
    std::cerr << "KokkosP: FATAL: Kokkos Tools Programming Interface's "
                 "tool-invoked Fence is NULL!\n";
    std::abort();
```

tests/sampler/test_parscan.cpp
```diff
@@ -0,0 +1,83 @@
...
+  std::cout << output.str() << std::endl;
+
+  //! Analyze test output.
+  for (const auto& matcher : matchers) {
+    EXPECT_THAT(output.str(), HasSubstr(matcher));
+  }
+
+  EXPECT_THAT(output.str(), Not(HasSubstr("KokkosP: sample 1 calling")));
+  EXPECT_THAT(output.str(), Not(HasSubstr("KokkosP: sample 75 calling")));
+  EXPECT_THAT(output.str(), Not(HasSubstr("KokkosP: sample 149 calling")));
```
masterleinad: I would rather do
```suggestion
   int occurrences = 0;
   while (std::string::size_type pos = 0; (pos = s.find("calling child-begin function", pos )) != std::string::npos) {
          ++occurrences;
          pos += target.length();
   }
   EXPECT_EQ(occurrences, 3);
```

tests/sampler/test_parscan.cpp
vlkale: Thanks for that. It is fixed.

tests/sampler/test_parscan.cpp
vlkale: Nice. But I think this should be two occurrences and not three, right? Specifically, I think you intended the parameter 3 in the function call in the last line to be 2. Only samples 51 and 102 are taken. 

And, why not have both your and my test in the code rather than replacing with yours? I think my code that you deleted is still covering different case (that a particular sample number is _not_ taken), so our two tests are complementary, right?

masterleinad: > Nice. But I think this should be two occurrences and not three, right? Specifically, I think you intended the parameter 3 in the function call in the last line to be 2. Only samples 51 and 102 are taken.

Sure.

> And, why not have both your and my test in the code rather than replacing with yours? I think my code that you deleted is still covering different case (that a particular sample number is _not_ taken), so our two tests are complementary, right?

I don't think we need both with what the tools are currently printing but it's fine with me if you want to have both.

vlkale:
> > > > Your test is really weak. It will pass if the sampler does nothing and just always forward to the underlying tool.
> 
> > > 
> 
> > > 
> 
> > > @dalg24 Thanks. However, I do not believe this is the case: It will not pass if the sampler does nothing. In my test code, _all_ matchers must be seen for this to pass. Also, the EXPECT_THAT function I have will cause a test to fail if the regular expression is not seen. This is the pertinent Google Tests documentation
> 
> > > https://google.github.io/googletest/reference/matchers.html
> 
> > 
> 
> > @dalg24 On second thought, you may be right - I re-read in the documentation that the matcher matches a single argument. The safest way to fix this is to just match the exact and full string I want to match in one matcher.
> 
> 
> 
> Assume we merge your test as it is proposed now and a follow PR includes this change.
> 
> ```diff
> 
> diff --git a/common/kokkos-sampler/kp_sampler_skip.cpp b/common/kokkos-sampler/kp_sampler_skip.cpp
> 
> index 78cbbff..86b55e3 100644
> 
> --- a/common/kokkos-sampler/kp_sampler_skip.cpp
> 
> +++ b/common/kokkos-sampler/kp_sampler_skip.cpp
> 
> @@ -193,7 +193,7 @@ void kokkosp_begin_parallel_for(const char* name, const uint32_t devID,
> 
>                                  uint64_t* kID) {
> 
>    *kID                          = uniqID++;
> 
>    static uint64_t invocationNum = 0;
> 
> -  ++invocationNum;
> 
> +  invocationNum;
> 
>    if ((invocationNum % kernelSampleSkip) == 0) {
> 
>      if (tool_verbosity > 0) {
> 
>        std::cout << "KokkosP: sample " << *kID
> 
> ```
> 
> Would you not want your test to fail?

I see you are showing a mistake by the programmer where they have not incremented the invocation number.  

The ideal way to test to cover that is to create a regular expression in RE2 that captures the fact that there exists _no_ samples between 0 and 149 that aren't 51 or 102.

However, as I was reading about RE2 last week in documentation and stackoverflow, there is a known problem with regular expression lookahead on some architectures and this would not be portable.

I think one easier way for now is to literally write an `EXPECT_THAT` for every single string that should not be in the output because sampling eliminated them, using Not(HasSubstr()). I had done that for three of them in my commit on Friday. Maybe even easier: reduce the number of invocations to 15, still taking two samples, and test the above by writing just 13 EXPECT_THAT's that test that those 13 samples haven't been taken. I will create this alternate simpler test to handle this. We can use this even easier one to test what you say.

I have a broader question here: What role should tests for Kokkos Tools play? 

In the code you show above, the tool connector programmer shouldn't have written that in the first place.

I think the tests for Kokkos Tools should emphasize more catching unexpected problems at runtime within the CI environments that are hard for tools connectors programmers to grasp without understanding Kokkos Tools and infrastructure. The null fence function pointer is just one example of this. The bug you bring up is important too, but it is another category of bugs involving a simpler programming error.

I think that testing of Kokkos Tools should follow testing in Kokkos core. A difference between Kokkos Tools and Kokkos core is that Kokkos core has more externally contributed libraries/connectors rather than contributions by Kokkos developers. That's why testing is very important in Kokkos Tools. I would say documentation for a Kokkos Tools connector developers (on how to write any Kokkos Tools connector) to complement testing Kokkos Tools may be a good thing to do.

dalg24:
> I have a broader question here: What role should tests for Kokkos Tools play?

Ensure that the tools behave as they are supposed to and are compatible with Kokkos Core.

If the tests we provide do not fail for the change I suggested above, then these tests are no good.
There are other ways to ensure that the tool output is right that do not involve complex regular expressions.

tests/CMakeLists.txt
vlkale: I don't know if that's right - I think there is a reason for  global fences in the sampler. Let me get you a better answer though. 

With that, can you clarify your statement 'kernels are empty'? You mean the kernel _callbacks_ are empty, right?

tests/sampler/test_parscan.cpp
vlkale: OK, got it. 

I will commit both (with the change of 2 occurrences for your suggested change) for now.

vlkale:
> 



> > I have a broader question here: What role should tests for Kokkos Tools play?
> 
> Ensure that the tools behave as they are supposed to and are compatible with Kokkos Core.
> 

OK. This is consistent with the way I see it at a high-level, and this captures it concisely. I will make sure the tests address these two aspects. 

By the way, I think Kokkos Tools connector documentation should also lay out this high-level requirement of testing for Kokkos Tools connectors developers so they can write out their own tests - this can be looked at later though. 

> If the tests we provide do not fail for the change I suggested above, then these tests are no good. 

That makes sense and sounds good. I guess you were showing that example for simplicity. Maybe another possibility is that the code is correct but the variable `invocationNum` gets overwritten due to memory corruption from system-level/OS events. Anyway, it makes sense to create a test that invocationNum is indeed getting incremented appropriately. 

> There are other ways to ensure that the tool output is right that do not involve complex regular expressions.

Yes, I will do without. A regex can simplify code and repetition, but it's looking like not here.

common/kokkos-sampler/kp_sampler_skip.cpp
vlkale: Yes, but one problem here is that I don't know if the current test will catch this if it is std:cerr. I have reverted back to std::cout for now. 

There may be a way in Gtests to capture cerr (I see EXPECT_DEATH but I don't think that solves this problem).  For simplicity and since this works right now, I am keeping this to std::cout.

tests/CMakeLists.txt
vlkale: To follow-up, see #117 where fences in the sampler are used in the same way I am doing. 

In terms why we need to test with fences enabled in the sampler, the kernels themselves (more specifically the kernel callbacks themselves) are printing because of the kernel-logger.

vlkale:
@dalg24 @masterleinad I think I have answered most of your questions. The CI tests are now passing with the changes. Please let me know if there is still anything unresolved. I have left all comments unresolved until you have looked at it. 

@masterleinad I am keeping std::cout for this PR for the fatal error, see comment above on why.

common/kokkos-sampler/kp_sampler_skip.cpp
vlkale: I have taken out the setting of global fences as you requested. 

(Also, I was getting confused on if you were saying to get rid of global fences in sampling altogether).

tests/CMakeLists.txt
vlkale: I have taken out the setting of global fences as you requested.

(Also, I was getting confused on if you were saying to get rid of global fences in sampling altogether).

crtrott (approved):

PR #249, opened 2024-04-04
kp_kernel_logger.cpp: typo in end scan call back function 

vlkale:
The end callback for scan in kp_kernel_logger.cpp has a typo in the function.  This never broke kernel logger and didn't not cause a problem unless one used parallel_scan in their code. This causes a parent tool or tool utility to not recognize that this end callback exists when the tool utility or parent tool checks that its child library function call exists. This impacts the sampler utility and can also impact the kernel filter utility. 

In any case, this one character edit to the code file needs to be fixed as soon as possible so that Kokkos user using Kokkos Tools kernel logger sees prints of parallel_scan in their kernel_logger output. 

This is part of the original PR #213 but I am breaking that one down into small changes since there are a sizeable number and this impacts more than just the sampler.

### DIFF
--- a/debugging/kernel-logger/kp_kernel_logger.cpp
+++ b/debugging/kernel-logger/kp_kernel_logger.cpp
@@ -100,7 +100,7 @@ extern "C" void kokkosp_begin_parallel_scan(const char* name,
   printf("    %s\n", name);
 }
 
-extern "C" void kokkospk_end_parallel_scan(const uint64_t kID) {
+extern "C" void kokkosp_end_parallel_scan(const uint64_t kID) {
   printf("KokkosP: Execution of kernel %llu is completed.\n",
          (unsigned long long)(kID));
 }


dalg24 (approved):
It would be a good time to add a unit test

vlkale:
> It would be a good time to add a unit test

Yes, that makes sense. Do you mean for this PR? I can do that - not a problem.

By the way, there needs to be a (possibly large - in any case systematic) effort to develop tests for each tool connector. Maybe we do that in a new PR and create a corresponding Issue.

vlkale:
@dalg24 Actually, I see you approved this so I am guessing you mean put a unit test for the kernel-logger in another PR. Yes, I will do that. It will be specifically a test for the kernel-logger in the tests subdirectory of Kokkos Tools.

masterleinad (approved):

PR #251, opened 2024-04-29
Remove KOKKOSTOOLS_LIBRARY_MODE

masterleinad:
`KOKKOSTOOLS_LIBRARY_MODE` isn't a CMake variable (and not documented) and there doesn't seem much of an advantage over using the implicit `BUILD_SHARED_LIBS` option. Note that `kp_add_library` expects all extra arguments to be source files anyway so specifying `KOKKOSTOOLS_LIBRARY_MODE` wouldn't actually work.


vlkale:
Looks OK to me especially if KOKKOSTOOLS_LIBRARY_MODE

is not doing anything anyway given what `kp_add_library` expects. Note that the sampler and filter are the only two utilities in Kokkos Tools, and I wonder why this was done just for Kokkos Tools.

In any case, I am good with removing it.

vlkale (approved):

crtrott (approved):

PR #252, opened 2024-04-29
Don't link against kokkostools in tests unconditionally

masterleinad:
Tests would normally just use environment variables to load the respective tool. In any case, we shouldn't link to `kokkostools` unconditionally. Tests that set the respective hooks explicitly would not use the environment variable `KOKKOS_TOOLS_LIBS` and only link against `kokkostools` that can either happen explicitly or via another argument to `kp_add_executable_and_test`.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -32,7 +32,7 @@ function(kp_add_executable_and_test)
     target_link_libraries(
         ${kaeat_args_TARGET_NAME}
         PRIVATE
-            kokkostools test_common
+            test_common
     )
 
     add_test(


vlkale:
Thanks @masterleinad 

Imposing kokkostools (monolithic) library requirement only for those Kokkos Tools tests needing it will make it easier for users to run the Kokkos Tools tests on their own platforms. 
I think none of the tests in develop need more than the individual (not monolithic) library for KOKKOS_TOOLS_LIBS.

vlkale (approved):
LGTM. 

Maybe we put a comment here  to developers of Kokkos Tools connectors to only use the monolithic library if it’s needed for a particular test (i.e., we explain the best practice that you conveyed in the description of the PR). 

It would be equally fine with me to put this in internal-documents on Kokkos Tools for developers.

crtrott (approved):

PR #257, opened 2024-06-06
Check for de/allocations pointing to null

### LINKED ISSUE #256: LAMMPS `develop` triggers assertion when `kokkosp_allocate_data` called with nullptr and size 0
https://github.com/kokkos/kokkos-tools/blob/58258bd6666dc6d11cb748441f62e68f9d22ba27/profiling/space-time-stack/kp_space_time_stack.cpp#L525

I don't know anything about LAMMPS, but I observe it triggering `kokkosp_allocate_data` with nullptr and size 0 allocation twice, which triggers this assertion. Interest in a PR that prints a warning in that specific case, but doesn't abort?

`patch_17Apr2024` release in LAMMPS also has this problem. `stable_2Aug2023_update3` does not trigger this problem.

@simongdg

vlkale:
Fix #256. 

To address this problem showing up in LAMMPS, the allocate and deallocate functions of the Kokkos Tool library space time stack (which record and maintain in a C++ set the memory allocations in a Kokkos application) need to have a check for the case of two allocations being null. The comparison operator (<) cannot distinguish two elements in the set that are both null. 

We need to have null pointer check of the parameter `ptr`, at the beginning of allocate and deallocate function.  If the ptr variable null, then the function should simply return. 

Within this null pointer check in space-time-stack allocation and deallocate functions, we should also discern whether the size of the ptr for allocation or deallocation is greater than 0 while the ptr is null. For that reason, we put in an assert just before returning.

### DIFF
--- a/profiling/space-time-stack/kp_space_time_stack.cpp
+++ b/profiling/space-time-stack/kp_space_time_stack.cpp
@@ -521,14 +521,20 @@ struct Allocations {
   Allocations() : total_size(0) {}
   void allocate(std::string&& name, const void* ptr, std::uint64_t size,
                 StackNode* frame) {
-    if(ptr == nullptr) { assert(size==0); return; }
+    if (ptr == nullptr) {
+      assert(size == 0);
+      return;
+    }
     auto res = alloc_set.emplace(Allocation(std::move(name), ptr, size, frame));
     assert(res.second);
     total_size += size;
   }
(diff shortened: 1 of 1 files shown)


masterleinad:
You need to run `clang-format`.

vlkale:
> You need to run `clang-format`.

Thanks! Done.

masterleinad (approved):
Fine with me.

crtrott (approved):

PR #258, opened 2024-06-12
Fix finding Variorum via environment variable

masterleinad:
Without this patch the hint `$ENV{VARIORUM_ROOT})` is ignored and overwritten by `${VARIORUM_ROOT}`.


vlkale:
LGTM.

vlkale (approved):

crtrott (approved):

dalg24:
The entire logic to find the package looks flawed. CMake will consider the _ROOT env variable by default and we should not be messing with _DIR variable

Issue #259, opened 2024-06-20
Perfetto results dependent on compile time setup


PhilipDeegan:
Hi there,

I'm trying to integrate these tools into a HPC code https://github.com/PHAREHUB/PHARE/

I've noticed that if compile with the following 

```
CMAKE_CXX_FLAGS="-DNDEBUG -g0 -O3 -march=native -mtune=native"
CMAKE_BUILD_TYPE="Release"
```

I get strange results and duplicate entries in the flame graph

I get expected results with the following:

```
CMAKE_CXX_FLAGS="-g3 -O3 -march=native -mtune=native -fno-omit-frame-pointer -pg"
CMAKE_BUILD_TYPE="RelWithDebInfo"
```

Is this expected, or is there some requirement that's not met in the first case?

PhilipDeegan:
on a closer inspection the latter config is still somewhat wrong, just *less* wrong

vlkale:
Thanks for pointing this out. Actually, we (specifically me, @uliegecsm and @crtrott) are aware of problems with perfetto when building it and are trying to fix. 

See Kokkos Tools PR #212 along with this branch of it: https://github.com/kokkos/kokkos-tools/compare/develop...cuda-11.0  that I am trying to merge into PR #212. 

I think the fixes of that PR may resolve the problems you are having. 

(FYI: We've decided right now we need to reach out to the developers of Perfetto to solve this, and it may take a week or two of turnaround time in talking with them to resolve that PR.)

PR #212, opened 2023-10-05
cicd: test Cuda version supported by Kokkos (11.0.3)

romintomasetti:
As requested by @crtrott in https://github.com/kokkos/kokkos-tools/pull/211#issuecomment-1749332578.

The closest to `11.0` in the Docker hub is `11.0.3`.

PR #262, opened 2024-07-11
CI test with CUDA 11.0 + Perfetto connector fix (bringing PR 212 to completion)

vlkale:
This PR is a continuation of PR #212, based on review and fixes of that PR and trying to address an associated problem /bug with Perfetto connector that is invoked when using CUDA-11.0, as discussed with @crtrott. 

Note that the problem is from perfetto/perfetto.cc and not the Perfetto Kokkos Tools connector. 

This PR also may address a Kokkos Tools Issue #259 from a Kokkos Tools user, but more discussion with the user is needed to see if it does fix it for their setup/environment/compilers that they are using for Perfetto.

### DIFF
--- a/.github/workflows/build-with-kokkos.yml
+++ b/.github/workflows/build-with-kokkos.yml
@@ -13,7 +13,7 @@ jobs:
           - image: ubuntu:22.04
             preset: OpenMP
             compiler: default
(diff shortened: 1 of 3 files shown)


crtrott:
Closed because part was merged rest with #271

PR #264, opened 2024-07-16
add "=" between --kokkos-tools-args and argument

cwpearson:
A user at LANL tried `--kokkos-tools-args 10` and received an error during `Kokkos::initialize()`. They needed to do `--kokkos-tools-args=10`

### DIFF
--- a/profiling/space-time-stack/kp_space_time_stack.cpp
+++ b/profiling/space-time-stack/kp_space_time_stack.cpp
@@ -913,9 +913,9 @@ Description:
 
 Example:
   The following example would set the threshold to 10%
-    <exe> [--kokkos-tools-args 10 ]
+    <exe> [--kokkos-tools-args=10 ]
 )usage";
-  std::cout << "usage: " << exe << "[--kokkos-tools-args <threshold>]\n"
+  std::cout << "usage: " << exe << "[--kokkos-tools-args=<threshold>]\n"
             << usage;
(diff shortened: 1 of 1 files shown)


dalg24:
Out of curiosity what error did we print?  (Wondering if that needs improvement too)

dalg24 (approved):

cwpearson:
I'm pretty sure it was this case:

https://github.com/kokkos/kokkos/blob/8e682582fd8302008ce5e422377e4ef2d730c23b/core/src/impl/Kokkos_Profiling.cpp#L127-L129

it was confusing because it just said

```
Warning: command line argument --kokkos-tools-args is not recognized
```

dalg24:
> I'm pretty sure it was this case:
> 
> https://github.com/kokkos/kokkos/blob/8e682582fd8302008ce5e422377e4ef2d730c23b/core/src/impl/Kokkos_Profiling.cpp#L127-L129
> 
> it was confusing because it just said
> 
> ```
> Warning: command line argument --kokkos-tools-args is not recognized
> ```

~~No way.  We explicitly guard with
https://github.com/kokkos/kokkos/blob/8e682582fd8302008ce5e422377e4ef2d730c23b/core/src/impl/Kokkos_Profiling.cpp#L125-L126~~
I spoke too fast.  OK we can definitely do better than that...

PR #266, opened 2024-07-27
Don't use `cmake_path` before CMake 3.20

cwpearson:
cmake_path in cmake/utils.cmake was added in CMake 3.20

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,7 +77,11 @@ endif()
 # Libraries
 if(KokkosTools_ENABLE_PAPI)
   find_package(PAPI REQUIRED) # TODO: papi-connector requires v6.0 or newer
-  cmake_path(GET PAPI_INCLUDE_DIR PARENT_PATH PAPI_ROOT)
+  if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.20")
+    cmake_path(GET PAPI_INCLUDE_DIR PARENT_PATH PAPI_ROOT)
+  else()
+    get_filename_component(PAPI_ROOT ${PAPI_INCLUDE_DIR} DIRECTORY)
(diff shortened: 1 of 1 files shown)


dalg24:
I would prefer if we did
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 4529918..198080c 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,7 +77,11 @@ endif()
 # Libraries
 if(KokkosTools_ENABLE_PAPI)
   find_package(PAPI REQUIRED) # TODO: papi-connector requires v6.0 or newer
-  cmake_path(GET PAPI_INCLUDE_DIR PARENT_PATH PAPI_ROOT)
+  if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.20")
+    cmake_path(GET PAPI_INCLUDE_DIR PARENT_PATH PAPI_ROOT)
+  else()
+    get_filename_component(PARENT_DIR(PARENT_DIR ${PAPI_INCLUDE_DIR} PAPI_ROOT)
+  endif()
   message(STATUS "Found PAPI ${PAPI_VERSION_STRING} at ${PAPI_ROOT}")
   set(KokkosTools_HAS_PAPI ON)
 else()
```
(not tested)
rather than forcing our hand to bump the minimum.  Printing information on the screen for an optional package should not be the thing dictating when we do so and what the minimum should be.

cwpearson:
I will test this with cmake 3.19 and 3.20 and see what happens

cwpearson:
Tested with PAPI 6 and cmake 3.19 / 3.20

dalg24 (approved):

masterleinad (approved):

PR #270, opened 2024-09-05
[WIP] Prevent perfetto-connector from breaking Kokkos Tools CI

vlkale:
Fix problem with chunk size that occurs with different compile time configurations. This is expected to fix Kokkos Tools Issue #259. 

This PR is part of an effort to breaks down PR #262 into two separate pieces. This piece focuses on the Perfetto connector fix, while the other is on allowing for Debian interactive mode and ensuring the latest versions of CMake.

### DIFF
--- a/profiling/perfetto-connector/perfetto/perfetto.cc
+++ b/profiling/perfetto-connector/perfetto/perfetto.cc
@@ -49830,7 +49830,7 @@ constexpr uint8_t kChunkNeedsPatching =
 }  // namespace.
 
 constexpr size_t TraceBuffer::ChunkRecord::kMaxSize;
-constexpr size_t TraceBuffer::InlineChunkHeaderSize = sizeof(ChunkRecord);
+const size_t TraceBuffer::InlineChunkHeaderSize = sizeof(ChunkRecord);
 
 // static
 std::unique_ptr<TraceBuffer> TraceBuffer::Create(size_t size_in_bytes,


masterleinad:
> Fix problem with chunk size that occurs with different compile time configurations. This is expected to fix Kokkos Tools Issue #259.

Which compiler configuration deosn't allow using `sizeof` in a `constexpr` context?

vlkale:
> > Fix problem with chunk size that occurs with different compile time configurations. This is expected to fix Kokkos Tools Issue #259.
> 
> Which compiler configuration deosn't allow using `sizeof` in a `constexpr` context?

@masterleinad  It's cuda-11.0.3 from ubuntu, and there could be others, given the error seen. Actually, this draft PR doesn't capture the full changes at the moment. Look at PR #262 to see the full changes. 

Details: 

1. I put this as draft but in process of trying refactor this draft PR to remove any use of CUDA 11.0.3 in the CI. I will update the title with NOTHING TO SEE HERE. Feel free to comment here, but better would be PR #262. 

2. Note that `constexpr` problem you refer to is not the only fix in #262 that is to come here - the primary problem with perfetto-connector is `BUILTIN_CLOCK`. 

4. The problem remaining for CUDA 11.0.3 is with nvcc_wrapper - see the CI test for CUDA 11.0.3 in #262. I think there is a problem involving the docker container - my goal is to make that a separate GitHub PR so that the other changes made in PR #262 can be merged.

crtrott (approved):

Issue #275, opened 2024-10-06
Compilation with Kokkos develop broken after Kokkos PR 7394


maartenarnst:
Compilation of `kokkos-tools` with `kokkos` develop is broken after 
- https://github.com/kokkos/kokkos/pull/7394

The reason seems to be that the `kokkos` PR moved the function `int_for_synchronization_reason` from `Kokkos_Profiling.hpp` to `Kokkos_Profiling_Interface.hpp`. However, this file `Kokkos_Profiling_Interface.hpp` is also in `kokkos-tools/profiling/all/impl`. And so because of the `kokkos` PR, the two files are no longer in sync.

When we now compile `kokkos-tools` with `kokkos` develop, we get a compilation error saying that the compiler can't find `int_for_synchronization_reason`. The compilation error arises when compiling the `kokkos-tools` tests. That's indeed where both `kokkos-tools` and `kokkos` are needed, and the two versions of the same header being out of sync can cause a problem. 

```
In file included from /opt/kokkos/include/impl/Kokkos_Tools.hpp:24,
                 from /opt/kokkos/include/Kokkos_HostSpace.hpp:37,
                 from /opt/kokkos/include/OpenMP/Kokkos_OpenMP.hpp:30,
                 from /opt/kokkos/include/decl/Kokkos_Declare_OPENMP.hpp:21,
                 from /opt/kokkos/include/KokkosCore_Config_DeclareBackend.hpp:22,
                 from /opt/kokkos/include/Kokkos_Core.hpp:45,
                 from /workspaces/kokkos/kokkos-tools/tests/UnitTestMain.cpp:3:
/opt/kokkos/include/impl/Kokkos_Profiling.hpp: In function 'void Kokkos::Tools::Experimental::Impl::profile_fence_event(const std::string&, Kokkos::Tools::Experimental::SpecialSynchronizationCases, const FencingFunctor&)':
/opt/kokkos/include/impl/Kokkos_Profiling.hpp:211:39: error: there are no arguments to 'int_for_synchronization_reason' that depend on a template parameter, so a declaration of 'int_for_synchronization_reason' must be available [-fpermissive]
  211 |       name, device_id_root<Space>() + int_for_synchronization_reason(reason),
```

The reason that the `kokkos-tools` version of `Kokkos_Profiling_Interface.hpp` has precedence over the `kokkos` version, and we thus get the error here, seems be that `kokkos-tools` uses hard `include_directories` rather than a `CMake` target:

[...]

PR #265, opened 2024-07-19
Variorium Connector: kernel granularity and JSON

twilk10:
Updates to the current connector with updates such as json updates, energy estimation, device information. It has also been altered to get power information at the kernel level.

PR #276, opened 2024-10-16
Kernel logger upgrades for better output and better code comparisons

csiefer2:
This PR does two things:

1) Upgrades the output on the Kernel logger to look more like the Tpetra DeepCopyTools --- rather than output space ID's we demanagle that and output names.

2) Allow you to use the KOKKOS_TOOLS_PROFILE_SUPPRESS_COUNTS variable to suppress the kernel counts in the output --- this is really useful if you're comparing two versions of the code with roughly the same flow, but some different kernel calls here and there.  Specifically, code paths with and without UnifiedMemory.  The call counting makes diff really hard and so turning it off can make diff easier.

### DIFF
--- a/debugging/kernel-logger/kp_kernel_logger.cpp
+++ b/debugging/kernel-logger/kp_kernel_logger.cpp
@@ -29,61 +29,71 @@ struct SpaceHandle {
   char name[64];
 };
 
-
-
-
 // Get a useful label from the deviceId
-// NOTE: Relevant code is in: kokkos/core/src/impl/Kokkos_Profiling_Interface.hpp
+// NOTE: Relevant code is in:
+// kokkos/core/src/impl/Kokkos_Profiling_Interface.hpp
 std::string deviceIdToString(const uint32_t deviceId) {
   using namespace Kokkos::Tools::Experimental;
   std::string device_label("(");
   ExecutionSpaceIdentifier eid = identifier_from_devid(deviceId);
-  if      (eid.type == DeviceType::Serial)       device_label+="Serial";
-  else if (eid.type == DeviceType::OpenMP)       device_label+="OpenMP";
-  else if (eid.type == DeviceType::Cuda)         device_label+="Cuda";
-  else if (eid.type == DeviceType::HIP)          device_label+="HIP";
-  else if (eid.type == DeviceType::OpenMPTarget) device_label+="OpenMPTarget";
-  else if (eid.type == DeviceType::HPX)          device_label+="HPX";
-  else if (eid.type == DeviceType::Threads)      device_label+="Threads";
-  else if (eid.type == DeviceType::SYCL)         device_label+="SYCL";
-  else if (eid.type == DeviceType::OpenACC)      device_label+="OpenACC";
-  else if (eid.type == DeviceType::Unknown)      device_label+="Unknown";
-  else                                           device_label+="Unknown to KokkosTools";
-  
-  if(eid.instance_id == int_for_synchronization_reason(SpecialSynchronizationCases::GlobalDeviceSynchronization))
+  if (eid.type == DeviceType::Serial)
+    device_label += "Serial";
+  else if (eid.type == DeviceType::OpenMP)
+    device_label += "OpenMP";
+  else if (eid.type == DeviceType::Cuda)
+    device_label += "Cuda";
+  else if (eid.type == DeviceType::HIP)
+    device_label += "HIP";
+  else if (eid.type == DeviceType::OpenMPTarget)
+    device_label += "OpenMPTarget";
+  else if (eid.type == DeviceType::HPX)
+    device_label += "HPX";
+  else if (eid.type == DeviceType::Threads)
+    device_label += "Threads";
+  else if (eid.type == DeviceType::SYCL)
+    device_label += "SYCL";
+  else if (eid.type == DeviceType::OpenACC)
+    device_label += "OpenACC";
+  else if (eid.type == DeviceType::Unknown)
+    device_label += "Unknown";
+  else
+    device_label += "Unknown to KokkosTools";
+
+  if (eid.instance_id ==
+      int_for_synchronization_reason(
+          SpecialSynchronizationCases::GlobalDeviceSynchronization))
     device_label += " All Instances)";
-  else if(eid.instance_id == int_for_synchronization_reason(SpecialSynchronizationCases::DeepCopyResourceSynchronization))
+  else if (eid.instance_id ==
+           int_for_synchronization_reason(
+               SpecialSynchronizationCases::DeepCopyResourceSynchronization))
     device_label += " DeepCopyResource)";
   else
     device_label += " Instance " + std::to_string(eid.instance_id) + ")";
-  
+
   return device_label;
 }
 
-
-
-
 bool suppressCounts() {
-  static bool value=false;
-  static bool initialized=false;
-  
-  if(initialized) 
+  static bool value       = false;
+  static bool initialized = false;
+
+  if (initialized)
     return value;
   else {
     const char* varVal = std::getenv("KOKKOS_PROFILE_SUPPRESS_COUNTS");
-    if(varVal) {
+    if (varVal) {
       std::string v = std::string(varVal);
       // default to false
-      if (v=="1" || v=="ON" || v=="on" || v=="TRUE" || v=="true" || v=="YES" || v=="yes") 
-	value=true;
+      if (v == "1" || v == "ON" || v == "on" || v == "TRUE" || v == "true" ||
+          v == "YES" || v == "yes")
+        value = true;
       else
-	value=false;
+        value = false;
     }
-    initialized=true;
+    initialized = true;
     return value;
   }
-} 
-
+}
 
 void kokkosp_print_region_stack_indent(const int level) {
   printf("KokkosP: ");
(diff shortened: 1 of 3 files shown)


csiefer2:
Yeah, yeah, I'll fix clang later.  Right now I want feedback.

vlkale:
@csiefer2 

The CI check formatting-check with clang-tidy is fine to not pass for now for the sake of a PR review. However, the CI check for simple-build is showing `error: ‘int_for_synchronization_reason’ was not declared in this scope`. It looks like the rest of the CI checks are also not passing, for the same reason.

Also, JM2C, but it may be good to separate this out into two PRs. I think the first change does not depend on whether the Kokkos Tools user sets `KOKKOS_PROFILE_SUPPRESS_COUNTS`, AFAIS. I also think the first change isn't controversial while the second change will need more time / thinking IMO: I wonder if the second change has a more efficient implementation given the conditionals in it. This is just a suggestion from my experience - feel free to keep it as is if you think it's better as one PR.

csiefer2:
@vlkale This is a debugging tool that fences all over the place. Host-side conditionals are basically irrelevant in terms of the performance hit the fences create.

As for the compile error, that's because the headers are no longer consistent with the headers in Kokkos.  I'll add a fix to that in a second.

csiefer2:
Compiles just fine now...

debugging/kernel-logger/kp_kernel_logger.cpp
```diff
@@ -19,14 +19,82 @@
...
+  else if (eid.type == DeviceType::Threads)
+    device_label += "Threads";
+  else if (eid.type == DeviceType::SYCL)
+    device_label += "SYCL";
+  else if (eid.type == DeviceType::OpenACC)
+    device_label += "OpenACC";
+  else if (eid.type == DeviceType::Unknown)
+    device_label += "Unknown";
+  else
+    device_label += "Unknown to KokkosTools";
```
masterleinad, line 60: I wouldn't be opposed to pushing this part (or even the whole function) to `Kokkos_Profiling_Interface.hpp`. Getting string out of the device id doesn't seem to be specific to the kernel logger tool (also see https://github.com/kokkos/kokkos-tools/pull/265/files#diff-839f34fcb31addd9a48252bebf4d37cf674f6fcdd16539c7039813192674b9c0R165-R187).

debugging/kernel-logger/kp_kernel_logger.cpp
```diff
@@ -19,14 +19,82 @@
...
+      // default to false
+      if (v == "1" || v == "ON" || v == "on" || v == "TRUE" || v == "true" ||
+          v == "YES" || v == "yes")
+        value = true;
+      else
+        value = false;
+    }
+    initialized = true;
+    return value;
+  }
```
masterleinad:
```suggestion
  static bool value       = [](){
    const char* varVal = std::getenv("KOKKOS_PROFILE_SUPPRESS_COUNTS");
    if (varVal) {
      std::string v = std::string(varVal);
      // default to false
      if (v == "1" || v == "ON" || v == "on" || v == "TRUE" || v == "true" ||
          v == "YES" || v == "yes")
        return true;
    }
    return false;
  }();
  return value;
```
to make it a little more concise.

masterleinad: Since `KOKKOS_PROFILE_LIBRARY` is deprecated, I would rather choose `KOKKOS_TOOLS_LOGGER_SUPPRESS_COUNTS` to also make clear that this *is* specific to the kernel logger tool.

csiefer2: We can do that as well. Lemme fix.

csiefer2: Done.  See below.

debugging/kernel-logger/kp_kernel_logger.cpp
csiefer2: Concur.  I also have to maintain this function in Trilinos, so I would *love* to put it in Kokkos_Profiling_Interface.hpp.

csiefer2: If @crtrott doesn't object, I'll open a Kokkos PR with that.

vlkale:
> @vlkale This is a debugging tool that fences all over the place. Host-side conditionals are basically irrelevant in terms of the performance hit the fences create.

True about fencing (aside: maybe in the future this tool could reduce fencing more, e.g., just allowing for global fencing to be turned off, if some Kokkos users want that). Stepping back though and forgetting about efficiency, I think I may be actually misunderstanding the purpose. I thought you are handling a case where the output of kernel-logger is not different across different Kokkos backends for a particular Kokkos application like LAMMPS, but also is non-deterministic for a given Kokkos backend. Did I get that right from "comparing two versions of the code with roughly the same flow, but some different kernel calls here and there. Specifically, code paths with and without UnifiedMemory."  ?

> 
> As for the compile error, that's because the headers are no longer consistent with the headers in Kokkos. I'll add a fix to that in a second.

Thanks and I realized just now that this is the issue that Kokkos Tools Github Issue #275 brings up.

csiefer2:
> I think I may be actually misunderstanding the purpose. I thought you are handling a case where the output of kernel-logger is not different across different Kokkos backends for a particular Kokkos application like LAMMPS, but also is non-deterministic for a given Kokkos backend. Did I get that right from "comparing two versions of the code with roughly the same flow, but some different kernel calls here and there. Specifically, code paths with and without UnifiedMemory." ?

I'm interested using the tool to track down cases in the case code flow is deterministic, but the answer is not due to an unintended race condition.

debugging/kernel-logger/kp_kernel_logger.cpp
```diff
@@ -19,14 +19,75 @@
...
 };
 
+// Get a useful label from the deviceId
+// NOTE: Relevant code is in:
+// kokkos/core/src/impl/Kokkos_Profiling_Interface.hpp
+std::string deviceIdToString(const uint32_t deviceId) {
+  using namespace Kokkos::Tools::Experimental;
+  std::string device_label("(");
+  ExecutionSpaceIdentifier eid = identifier_from_devid(deviceId);
+  if (eid.type == DeviceType::Serial)
```
romintomasetti, line 39: This is a switch case.

romintomasetti (changes requested):

debugging/kernel-logger/kp_kernel_logger.cpp
```diff
@@ -19,14 +19,75 @@
...
+  return device_label;
+}
+
+bool suppressCounts() {
+  static bool value       = [](){
+    const char* varVal = std::getenv("KOKKOS_TOOLS_LOGGER_SUPPRESS_COUNTS");
+    if (varVal) {
+      std::string v = std::string(varVal);
+      // default to false
+      if (v == "1" || v == "ON" || v == "on" || v == "TRUE" || v == "true" ||
```
romintomasetti, line 82: Most of the time, it seems that this repo wants either 0 or 1 for the value of an environment variable.

Then use `atoi`.

I think it is important that all environment variables follow consistent conventions.

debugging/kernel-logger/kp_kernel_logger.cpp
```diff
@@ -19,14 +19,75 @@
...
+    device_label += " DeepCopyResource)";
+  else
+    device_label += " Instance " + std::to_string(eid.instance_id) + ")";
+
+  return device_label;
+}
+
+bool suppressCounts() {
+  static bool value       = [](){
+    const char* varVal = std::getenv("KOKKOS_TOOLS_LOGGER_SUPPRESS_COUNTS");
```
romintomasetti, line 78: Please update your PR description in which you're currently mentioning `KOKKOS_PROFILE_SUPPRESS_COUNTS`.

debugging/kernel-logger/kp_kernel_logger.cpp
```diff
@@ -80,19 +143,23 @@ extern "C" void kokkosp_begin_parallel_for(const char* name,
...
+  *kID       = uniqID++;
+  int output = *kID;
+  if (suppressCounts()) output = 0;
 
   printf(
-      "KokkosP: Executing parallel-scan kernel on device %d with unique "
+      "KokkosP: Executing parallel-scan kernel on device %s with unique "
       "execution identifier %llu\n",
-      devID, (unsigned long long)(*kID));
+      deviceIdToString(devID).c_str(), (unsigned long long)(output));
```
romintomasetti: Wait... What if we have several devices ? For instance, several `Cuda` GPUs.

Seems this change is really not going in the direction of allowing devices to be distinguished.

I would be OKish with:
```suggestion
      printf(
      "KokkosP: Executing parallel-scan kernel on device %d (%s) with unique "
      "execution identifier %llu\n",
      devID, deviceIdToString(devID).c_str(), (unsigned long long)(output));
```

romintomasetti:
```suggestion
      deviceIdToString(devID).c_str(), suppressCounts() ? 0 : *kID);
```

debugging/kernel-logger/kp_kernel_logger.cpp
csiefer2: @romintomasetti This is a Tpetra-ism to allow it to take anything reasonable as an option.  If @vlkale or @crtrott doesn't like it that's fine, but we've found the flexibility to be useful in Trilinos

debugging/kernel-logger/kp_kernel_logger.cpp
csiefer2: Done

janciesko:
@csiefer2 What's the status of this please?

PR #278, opened 2024-12-16
feat(roctxconnector): link `rocprofiler-sdk-roctx` as of `ROCm` 6.2

maartenarnst:
This PR updates the `CMake` logic for `roctxconnector` so as to follow AMD's recommendation to link with `rocprofiler-sdk-roctx` as of `ROCm` 6.2.

For reference, the recommendation is in the "note" on this webpage: 
- https://rocm.docs.amd.com/projects/rocprofiler-sdk/en/amd-mainline/how-to/using-rocprofv3.html

Note that with this PR, we'll use the "new" `rocprofiler-sdk-roctx` instead of the "old" one when it is available. Another possibility would be to have an option de determine which one is linked.

### DIFF
--- a/.github/workflows/build-with-kokkos.yml
+++ b/.github/workflows/build-with-kokkos.yml
@@ -22,7 +22,7 @@ jobs:
           - image: rocm/dev-ubuntu-22.04:5.4
             preset: ROCm
             compiler: default
-          - image: rocm/dev-ubuntu-22.04:5.7
+          - image: rocm/dev-ubuntu-22.04:6.3
             preset: ROCm
             compiler: default
     container:
--- a/profiling/roctx-connector/CMakeLists.txt
+++ b/profiling/roctx-connector/CMakeLists.txt
@@ -1,7 +1,16 @@
-find_library(ROCM_ROCTX_LIB roctx64 REQUIRED HINTS $ENV{ROCM_PATH}/lib)
-find_path(ROCM_ROCTX_INCLUDE roctx.h REQUIRED HINTS $ENV{ROCM_PATH}/include/roctracer)
-
 kp_add_library(kp_roctx_connector kp_roctx_connector.cpp)
 
-target_include_directories(kp_roctx_connector PRIVATE ${ROCM_ROCTX_INCLUDE})
-target_link_libraries(kp_roctx_connector PRIVATE ${ROCM_ROCTX_LIB})
+# As of ROCm 6.2, it is recommended to use ROCtx provided by rocprofiler-sdk-roctx
+# instead of the "old" one provided by roctracer.
+#
+# See also: https://rocm.docs.amd.com/projects/rocprofiler-sdk/en/amd-mainline/how-to/using-rocprofv3.html
+find_package(rocprofiler-sdk-roctx CONFIG PATHS $ENV{ROCM_PATH})
+if(rocprofiler-sdk-roctx_FOUND)
+    target_link_libraries(kp_roctx_connector PRIVATE rocprofiler-sdk-roctx::rocprofiler-sdk-roctx)
+else()
+    find_library(ROCM_ROCTX_LIB roctx64 REQUIRED HINTS $ENV{ROCM_PATH}/lib)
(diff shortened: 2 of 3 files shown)


profiling/roctx-connector/CMakeLists.txt
```diff
@@ -1,7 +1,18 @@
...
-
 kp_add_library(kp_roctx_connector kp_roctx_connector.cpp)
 
-target_include_directories(kp_roctx_connector PRIVATE ${ROCM_ROCTX_INCLUDE})
-target_link_libraries(kp_roctx_connector PRIVATE ${ROCM_ROCTX_LIB})
+# As of ROCm 6.2, it is recommended to use ROCtx provided by rocprofiler-sdk-roctx
+# instead of the "old" one provided by roctracer.
+#
+# See also: https://rocm.docs.amd.com/projects/rocprofiler-sdk/en/amd-mainline/how-to/using-rocprofv3.html
+find_package(rocprofiler-sdk-roctx CONFIG PATHS $ENV{ROCM_PATH} PATH_SUFFIXES "lib/cmake/rocprofiler-sdk-roctx")
```
dalg24: Why are we specifying `PATH_SUFFIXES`?  I expect that path will be constructed by CMake regardless.

dalg24: Also, why do you think we should add `ROCM_PATH` to the search paths?

profiling/roctx-connector/kp_roctx_connector.cpp
```diff
@@ -14,7 +14,11 @@
 //
 //@HEADER
 
-#include <roctx.h>
+#ifdef KP_ROCTXCONNECTOR_HAVE_ROCPROFILER_SDK_ROCTX
```
dalg24: Why do we need to define our own macro?
Does ROCprofiler-SDK not define anything we could leverage?
Could't we just use an `__has_include` expression?

.github/workflows/build-with-kokkos.yml
```diff
@@ -19,10 +19,10 @@ jobs:
           - image: nvidia/cuda:12.2.0-devel-ubuntu22.04
             preset: Cuda
             compiler: {cpp: g++-12, c: gcc-12}
-          - image: rocm/dev-ubuntu-22.04:5.4
+          - image: rocm/dev-ubuntu-22.04:5.7
```
dalg24: Why do we want to bump that version?

dalg24:
@Rombur plz review

profiling/roctx-connector/CMakeLists.txt
maartenarnst: OK. The `PATH_SUFFIXES` shouldn't be needed indeed. I still have to check for the `ROCM_PATH`.

profiling/roctx-connector/kp_roctx_connector.cpp
maartenarnst: OK. The `CMake` `config` provided by `rocprofiler-sdk-roctx` doesn't seem to set a compiler definition. I've followed your suggestion of using `__has_include`.

.github/workflows/build-with-kokkos.yml
maartenarnst: Just a habit of removing the oldest version in such a case :) We can keep the 5.4 release too. Just noting that on LUMI (the European supercomputer ~ Frontier), the ROCm 5.4 stack is no longer available.

Rombur: It is available on Frontier, so we should keep it.

profiling/roctx-connector/CMakeLists.txt
maartenarnst: I've just tested the need for the `ROCM_PATH`. If we don't specify it as a search path, then the package isn't found.

.github/workflows/build-with-kokkos.yml
maartenarnst: OK. No problem. Let's keep 5.4.

Rombur (approved):

dalg24 (approved):

PR #281, opened 2025-01-16
disable warnings in third-party code

cwpearson:
This is third-party code, so modify the compiler flags instead of the code itself.

### DIFF
--- a/profiling/perfetto-connector/CMakeLists.txt
+++ b/profiling/perfetto-connector/CMakeLists.txt
@@ -1,3 +1,7 @@
 kp_add_library(kp_perfetto_connector libperfetto-connector.cpp perfetto/perfetto.cc)
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    # this is third-party code, so we just disable some errors that pop up
+    target_compile_options(kp_perfetto_connector PRIVATE -Wno-error=redundant-move)
+endif()
 target_include_directories(kp_perfetto_connector PRIVATE perfetto)
 
--- a/profiling/perfetto-connector/CMakeLists.txt
+++ b/profiling/perfetto-connector/CMakeLists.txt
@@ -1,7 +1,14 @@
-kp_add_library(kp_perfetto_connector libperfetto-connector.cpp perfetto/perfetto.cc)
+# perfetto/perfetto.cc is third-party code, so we want to disable all warnings for it
+add_library(perfetto_static STATIC perfetto/perfetto.cc)
+set_target_properties(perfetto_static PROPERTIES POSITION_INDEPENDENT_CODE ON)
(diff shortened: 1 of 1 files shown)


profiling/perfetto-connector/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
 kp_add_library(kp_perfetto_connector libperfetto-connector.cpp perfetto/perfetto.cc)
+# this is third-party code, so we just disable some errors that pop up
+target_compile_options(kp_perfetto_connector PRIVATE -Wno-error=redundant-move)
```
dalg24: Is that portable to MSVC?

cwpearson:
`-Wno-error=redundant-move` was introduced in GCC 9.1 and Clang 3.7

masterleinad (approved):

dalg24 (commented):
If anything we should disable all warnings.
On my cell now but I am sure you can find examples of us doing that, possibly for gtest in core?

dalg24:
> If anything we should disable all warnings.
> On my cell now but I am sure you can find examples of us doing that, possibly for gtest in core?

We do it in perf_test for google benchmark. Sorry can't get a reference but I am sure you'll find it

cwpearson:
Cool, TIL about `-w` / `/w`.

Changed perfetto.cc to be compiled into its own static library so we still get warnings from our connector code.

profiling/perfetto-connector/CMakeLists.txt
```diff
@@ -1,3 +1,14 @@
-kp_add_library(kp_perfetto_connector libperfetto-connector.cpp perfetto/perfetto.cc)
-target_include_directories(kp_perfetto_connector PRIVATE perfetto)
+# perfetto/perfetto.cc is third-party code, so we want to disable all warnings for it
+add_library(perfetto_static STATIC perfetto/perfetto.cc)
+set_target_properties(perfetto_static PROPERTIES POSITION_INDEPENDENT_CODE ON)
```
masterleinad, line 3: Why do we built the library explicitly as `static` with `fpic` instead of shared?

masterleinad:
> Changed perfetto.cc to be compiled into its own static library so we still get warnings from our connector code.

Can you elaborate on that?

profiling/perfetto-connector/CMakeLists.txt
cwpearson: I figured it was less opportunity for something to go wrong during the build. I don't have a strong opinion.

cwpearson:
Used to be that our connector code (libperfetto-connector.cc) and the third-party perfetto.cc were compiled into the same target, so whatever options were applied to one would be applied to both. I split them up so we could ignore warnings from perfetto.cc and otherwise leave our code compiling the way it used to be.

masterleinad (approved):
Fine with me.

cwpearson:
I don't have the merge power, someone else will have to hit the button

masterleinad:
> I don't have the merge power, someone else will have to hit the button

Probably only @dalg24 and @crtrott.

masterleinad:
Fixed differently in #286.

PR #284, opened 2025-04-29
SpaceTimeStack: Account for MPI imbalance when applying printing threshold and add test

vbrunini:
So that entries with a low average time but high imbalance are reported since they are often the cause of significant performance problems due to load imbalance.

Also fix missing deep_copy hook exposure.


vlkale:
> …shold.
> 
> 
> 
> So that entries with a low average time but high imbalance are reported since they are often the cause of significant performance problems due to load imbalance.
> 
> 
> 
> Also fix missing deep_copy hook exposure.

@vbrunini 

Thanks for this. I am curious if you can provide an example code with the problem here in the PR description. Even better would be a set of tests via, say, Google Test, to check that those entries with low average time but high imbalance are indeed being reported.

A side note is that the fix for the missing deep_copy hook you have should ideally be a separate PR, as it looks conceptually unrelated to the change related to MPI load imbalance?

Thanks!

vbrunini:
> > …shold.
> > So that entries with a low average time but high imbalance are reported since they are often the cause of significant performance problems due to load imbalance.
> > Also fix missing deep_copy hook exposure.
> 
> @vbrunini
> 
> Thanks for this. I am curious if you can provide an example code with the problem here in the PR description. Even better would be a set of tests via, say, Google Test, to check that those entries with low average time but high imbalance are indeed being reported.
> 
> A side note is that the fix for the missing deep_copy hook you have should ideally be a separate PR, as it looks conceptually unrelated to the change related to MPI load imbalance?
> 
> Thanks!

Added a test (and split off the deep copy parts into #285)

tests/space-time-stack/test_deep_copy.cpp.orig
masterleinad, line 1: I'm pretty sure you don't want to keep this file.

profiling/space-time-stack/kp_space_time_stack.cpp
```diff
@@ -336,9 +336,11 @@ struct StackNode {
   void print_recursive(std::ostream& os, std::string my_indent,
                        std::string const& child_indent,
                        double tree_time) const {
-    auto percent = (total_runtime / tree_time) * 100.0;
+    const double comm_size = total_runtime / avg_runtime;
+    auto threshold_percent = ((max_runtime * comm_size) / tree_time) * 100.0;
+    auto percent           = (total_runtime / tree_time) * 100.0;
```
masterleinad: Can you elaborate on why this is sensible? I understand the previous code but now we essentially do
```
((max_runtime / avg_runtime) * percent  < outout_threshold
```
In other words, why does multiplying by `max_runtime / avg_runtime` do what you want? Wouldn't you want to do something like
```
if (percent > first_threshold || (high_imbalance && percent > second_threshold))
  report
```
? It's not quite clear to me that this second threshold should be the same as the orignal one divided by the imbalance. Maybe it's sensible to report all kernels that have an imbalance greater than a certain factor?

vbrunini: The thinking is to identify any kernels/regions that would be above the 0.1% threshold if we only looked at the runtimes on that rank. Assuming that total_runtime is evenly distributed across ranks (which I think is basically always true), then `rank_wall_time = tree_time / comm_size` and therefore `max_runtime / rank_wall_time == (max_runtime * comm_size) / tree_time`.

masterleinad: OK. You basically want to use `max_runtime` instead of `avg_runtime` for comparing with the threshold value. I think that's sensible. Not sure if we need a configuration option.
Note that there also is `print_json_recursive` that needs consistent logic.

profiling/space-time-stack/kp_space_time_stack.cpp
```diff
@@ -336,9 +336,11 @@ struct StackNode {
   void print_recursive(std::ostream& os, std::string my_indent,
                        std::string const& child_indent,
                        double tree_time) const {
-    auto percent = (total_runtime / tree_time) * 100.0;
+    const double comm_size = total_runtime / avg_runtime;
```
masterleinad: You might as well compute this properly once and store it in a member variable.

vbrunini:
Any suggestions for what to do about the CI failure from the tests running as root?
```
--------------------------------------------------------------------------
mpirun has detected an attempt to run as root.

Running as root is *strongly* discouraged as any mistake (e.g., in
defining TMPDIR) or bug can result in catastrophic damage to the OS
file system, leaving your system in an unusable state.

We strongly suggest that you run mpirun as a non-root user.

You can override this protection by adding the --allow-run-as-root option
to the cmd line or by setting two environment variables in the following way:
the variable OMPI_ALLOW_RUN_AS_ROOT=1 to indicate the desire to override this
protection, and OMPI_ALLOW_RUN_AS_ROOT_CONFIRM=1 to confirm the choice and
add one more layer of certainty that you want to do so.
We reiterate our advice against doing so - please proceed at your own risk.
--------------------------------------------------------------------------
```

tests/CMakeLists.txt
```diff
@@ -24,6 +25,15 @@ function(kp_add_executable_and_test)
...
 
+    if(DEFINED kaeat_args_RANKS AND NOT KokkosTools_ENABLE_MPI)
+        message(WARNING "Skipping ${kaeat_args_TARGET_NAME} test because MPI is not enabled.")
+        return()
+    endif()
+
+    if(DEFINED kaeat_args_RANKS)
+        message(WARNING "Test ${kaeat_args_TARGET_NAME} will run on ${kaeat_args_RANKS} MPI ranks.")
+    endif()
+
```
masterleinad: Not sure we need this.
```suggestion
```

tests/space-time-stack/test_threshold_with_imbalance.cpp
```diff
@@ -0,0 +1,81 @@
+#include <chrono>
```
janciesko, line 4:
```suggestion
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
// SPDX-FileCopyrightText: Copyright Contributors to the Kokkos project

#include <chrono>
```

tests/CMakeLists.txt
```diff
@@ -37,10 +43,18 @@ function(kp_add_executable_and_test)
...
     )
 
-    add_test(
-        NAME ${kaeat_args_TARGET_NAME}
-        COMMAND $<TARGET_FILE:${kaeat_args_TARGET_NAME}>
-    )
+    if(DEFINED kaeat_args_RANKS)
+        add_test(
+            NAME ${kaeat_args_TARGET_NAME}
+            COMMAND mpirun -n ${kaeat_args_RANKS} $<TARGET_FILE:${kaeat_args_TARGET_NAME}>
```
janciesko:
```suggestion
            COMMAND ${MPIEXEC_EXECUTABLE} -n ${kaeat_args_RANKS} $<TARGET_FILE:${kaeat_args_TARGET_NAME}>
```

tests/space-time-stack/test_threshold_with_imbalance.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+  //! Initialize @c Kokkos.
+  Kokkos::initialize();
+
+  //! Redirect output for later analysis.
+  std::cout.flush();
+  std::ostringstream output;
+  std::streambuf* coutbuf = std::cout.rdbuf(output.rdbuf());
+
+  //! Run tests. @todo Replace this with Google Test.
```
janciesko:
```suggestion
```

tests/space-time-stack/test_threshold_with_imbalance.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+static const std::vector<std::string> matchers{
+    "[0-9.e]+ sec [0-9.]+% [0-9.]+% 0.[0-9]+% 100.0% 0.00e\\+00 1 "
+    "all_ranks_region \\[region\\]",
+    "[0-9.e]+ sec [0-9.]+% [0-9.]+% [1-9][0-9.]+% 100.0% 0.00e\\+00 1 "
+    "rank_1_region \\[region\\]",
+};
+
+/**
+ * @test This test checks that the tool outputs deep_copy statistics.
```
janciesko:
```suggestion
 * @test This test checks that the tool reports imbalanced MPI regions that would be filtered out when using average time alone.
```

janciesko (changes requested):

janciesko:
@vbrunini Please rebase

janciesko (approved):

masterleinad (approved):

crtrott (approved):

PR #286, opened 2025-04-29
Perfetto remove redundant std::move

masterleinad:
The CI complains about a redundant `std::move`, see https://github.com/kokkos/kokkos-tools/actions/runs/14736547964/job/41363971558?pr=285.

### DIFF
--- a/profiling/perfetto-connector/perfetto/perfetto.cc
+++ b/profiling/perfetto-connector/perfetto/perfetto.cc
@@ -72862,7 +72862,7 @@ void ProducerIPCClientImpl::OnConnect() {
 
   // If there are pending Sync() requests, send them now.
   for (const auto& pending_sync : pending_sync_reqs_)
-    Sync(std::move(pending_sync));
+    Sync(pending_sync);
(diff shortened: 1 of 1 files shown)


profiling/perfetto-connector/perfetto/perfetto.cc
```diff
@@ -72862,7 +72862,7 @@ void ProducerIPCClientImpl::OnConnect() {
 
   // If there are pending Sync() requests, send them now.
   for (const auto& pending_sync : pending_sync_reqs_)
-    Sync(std::move(pending_sync));
+    Sync(pending_sync);
```
dalg24: I would have dropped the `const`-qualifier in the range-based for loop init-statement instead.

I don't think it matters too much but I checked how upstream perfetto resolved this and they did exactly that.
https://github.com/google/perfetto/blob/34953f0ef39e5496fcf9f4dd34e67de6b8dd100f/src/tracing/ipc/producer/producer_ipc_client_impl.cc#L233-L234

dalg24 (approved):

PR #288, opened 2025-05-06
Error out if the set CXX_COMPILER is not the same as Kokkos_CXX_COMPILER

### LINKED ISSUE #274: Option `KokkosTools_REUSE_KOKKOS_COMPILER` does not work
Because the call to `project` can be expected to already set `CMAKE_CXX_COMPILER`

- https://github.com/kokkos/kokkos-tools/blob/a9453b21e09c7555096f3a96bb3fa8b34afe19f8/CMakeLists.txt#L3

the option `KokkosTools_REUSE_KOKKOS_COMPILER` can no longer change `CMAKE_CXX_COMPILER`

- https://github.com/kokkos/kokkos-tools/blob/a9453b21e09c7555096f3a96bb3fa8b34afe19f8/CMakeLists.txt#L65-L68

so the option `KokkosTools_REUSE_KOKKOS_COMPILER` can be expected not to work.

Ways forward could be
- remove the option `KokkosTools_REUSE_KOKKOS_COMPILER`
- use `FORCE` in the `set` command

Using `FORCE` would indeed "reuse" the compiler, but it may not be a good solution because some flags may have been set already in the toolchain for the compiler that is now being overridden with force. 

Note that the `CMake` documentation says that `CMAKE_CXX_COMPILER` may be set only once:

- https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER.html

JBludau:
potentially fixes https://github.com/kokkos/kokkos-tools/issues/274 by erroring out. Rename the option to `KokkosTools_CHECK_CXX_COMPILER` and errors out if check fails.

Not implemented other solutions:
Another solution would be using `FORCE` in the `SET`.
Another solution would be deprecating the old option and just pushing this into the documentation.
Another solution would be to link to Kokkos::kokkos and have our build system do the work but linking to a library that we don't use


JBludau:
if https://github.com/kokkos/kokkos/pull/8138 gets merged this should become obsolete since `find_package(Kokkos)` will include a compatibility check

PR #289, opened 2025-05-08
Updates for ease of building

OlekRaymond:
I recently built kokkos tools on windows and found that some things could be communicated better, particularly why some features were not made available.

I have added warnings and more verbose messages to the CMake to make this more obvious.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,7 +31,8 @@ message(STATUS)
 # Common settings
 set(BUILD_SHARED_LIBS "Build shared libraries" ON)
 if(WIN32)
-  set(BUILD_SHARED_LIBS OFF) # We need to add __declspec(dllexport/dllimport) for Windows DLLs
+  set(BUILD_SHARED_LIBS OFF) 
+  message(STATUS  "We need to add __declspec(dllexport/dllimport) for Windows DLLs, until then windows building is not supported")
 endif()
 
 # Tools settings
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@ if(NOT DEFINED CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 17)
 endif()
 if(CMAKE_CXX_STANDARD LESS 17)
-  message(FATAL_ERROR "KokkosTools requires C++17")
+  message(FATAL_ERROR "KokkosTools requires C++17 or higher")
 endif()
 
 
@@ -19,13 +19,14 @@ cmake_policy(SET CMP0111 NEW) # error if library not found
 
 # Disable in-source builds to prevent source tree corruption.
 if(CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR)
-  message(FATAL_ERROR "FATAL: In-source builds are not allowed. You should create a separate directory for build files.")
+  message(FATAL_ERROR "FATAL: In-source builds are not allowed. You should create a separate directory for build files. \
+  This can be done with the invocation cmake -B build -S . to create a build directory called build")
 endif()
 
 list(INSERT CMAKE_MODULE_PATH 0 ${PROJECT_SOURCE_DIR}/cmake)
 
 message(STATUS)
-message(STATUS Configuring Kokkos-Tools)
+message(STATUS "Configuring Kokkos-Tools")
 message(STATUS)
 
 # Common settings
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -19,26 +19,25 @@ cmake_policy(SET CMP0111 NEW) # error if library not found
...
-message(STATUS Configuring Kokkos-Tools)
+message(STATUS "Configuring Kokkos-Tools")
 message(STATUS)
 
 # Common settings
 set(BUILD_SHARED_LIBS "Build shared libraries" ON)
 if(WIN32)
-  set(BUILD_SHARED_LIBS OFF) # We need to add __declspec(dllexport/dllimport) for Windows DLLs
+  set(BUILD_SHARED_LIBS OFF) 
+  message(STATUS  "We need to add __declspec(dllexport/dllimport) for Windows DLLs, until then windows building is not supported")
```
JBludau: This is not really a STATUS. Is there a way to add this to the library?

CMakeLists.txt
```diff
@@ -19,26 +19,25 @@ cmake_policy(SET CMP0111 NEW) # error if library not found
...
 endif()
 
 # Tools settings
-option(KokkosTools_ENABLE_SINGLE  "Build single library interfacing all profilers and dispatching at runtime" OFF)
-if(WIN32)
-  set(KokkosTools_ENABLE_SINGLE ON)
-endif()
+option(KokkosTools_ENABLE_SINGLE  "Build single library interfacing all profilers and dispatching at runtime" ${WIN32})
```
JBludau: I like the old approach better since it is easier to read and reason about

CMakeLists.txt
```diff
@@ -119,9 +118,6 @@ endif()
 # make Kokkos profiling interface available for native profilers
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/profiling/all)
 
-set(COMMON_HEADERS_PATH ${CMAKE_CURRENT_BINARY_DIR}/common)
-include_directories(${COMMON_HEADERS_PATH})
-
```
JBludau, line 111: why delete these?

CMakeLists.txt
```diff
@@ -139,7 +135,7 @@ if(WIN32)
 endif()
 
 if(APPLE)
-  message(STATUS "Apple OSX target detected.")
+  message(STATUS "Apple OSX target detected - skipping Unix-only tools.")
```
JBludau: it is misleading to print this here ... as this line is not skipping anything

profiling/nvtx-connector/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
 find_package(CUDAToolkit REQUIRED)
 kp_add_library(kp_nvtx_connector kp_nvtx_connector.cpp)
 
-target_link_libraries(kp_nvtx_connector CUDA::nvToolsExt)
+target_link_libraries(kp_nvtx_connector CUDA::nvtx3)
```
JBludau: how is that related to the description of the pull request?

CMakeLists.txt
```diff
@@ -19,26 +19,25 @@ cmake_policy(SET CMP0111 NEW) # error if library not found
 
 # Disable in-source builds to prevent source tree corruption.
 if(CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR)
-  message(FATAL_ERROR "FATAL: In-source builds are not allowed. You should create a separate directory for build files.")
+  message(FATAL_ERROR "FATAL: In-source builds are not allowed. You should create a separate directory for build files. \
+  This can be done with the invocation cmake -B build -S . to create a build directory called build")
```
JBludau: I would refrain from details how to do stuff in cmake ...

JBludau (changes requested):
this seems a bit over the place ... would it not be better to update documentation instead of the cmake?

CMakeLists.txt
```diff
@@ -25,20 +25,19 @@ endif()
...
 message(STATUS)
-message(STATUS Configuring Kokkos-Tools)
+message(STATUS "Configuring Kokkos-Tools")
 message(STATUS)
 
 # Common settings
-set(BUILD_SHARED_LIBS "Build shared libraries" ON)
-if(WIN32)
-  set(BUILD_SHARED_LIBS OFF) # We need to add __declspec(dllexport/dllimport) for Windows DLLs
+option(BUILD_SHARED_LIBS "Build shared libraries" ON)
```
JBludau, line 32: This is not really an option. The tools will only work as shared libs.

CMakeLists.txt
```diff
@@ -25,20 +25,19 @@ endif()
...
+# We need to add __declspec(dllexport/dllimport) for Windows DLLs
+if(WIN32 AND BUILD_SHARED_LIBS)
+  set(BUILD_SHARED_LIBS OFF) 
+  message(STATUS "Windows building of .dlls is not yet supported")
```
JBludau, line 36: I think we should error out

CMakeLists.txt
OlekRaymond: > This is not really a STATUS 

I agree, I wanted to keep the things that were being built on windows remaining for pipelines and testing.

> Is there a way to add this to the library?

I am intending to do so under another ticket

OlekRaymond:
Replaced by pull request #298

I had a think about what the best thing to do is.

It would make sense to leave anything that can compile on Windows to remain able to compile so it can be checked on pipelines.

However there are no pipelines that compile anything on Windows so that's all moot. 

I am going to open a new PR that fails early if on windows (and a cmake variable is false so it can be turned off for dev) with a message explaining it's not supported.

PR #290, opened 2025-05-12
update(`nvtx-connector`): use `nvtx3`

maartenarnst:
Nvidia has deleted `nvtx` `v2` in `CUDA` Toolkit  12.9:

- https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#id4

Following their recommendation, this PR updates the `nvtx-connector` to use `nvtx3`.

- https://github.com/NVIDIA/NVTX

Note that we have checked that `nvtx3` is available in `Docker` images dating back to at least `Cuda` 12.1.

### DIFF
--- a/profiling/nvtx-connector/CMakeLists.txt
+++ b/profiling/nvtx-connector/CMakeLists.txt
@@ -1,4 +1,11 @@
 find_package(CUDAToolkit REQUIRED)
 kp_add_library(kp_nvtx_connector kp_nvtx_connector.cpp)
 
-target_link_libraries(kp_nvtx_connector CUDA::nvToolsExt)
+# Fallback since the imported target CUDA::nvtx3 is defined only as of CMake 3.25.
+if(NOT TARGET CUDA::nvtx3)
+    add_library(CUDA::nvtx3 INTERFACE IMPORTED)
+    target_include_directories(CUDA::nvtx3 INTERFACE "${CUDAToolkit_INCLUDE_DIRS}")
+    target_link_libraries(CUDA::nvtx3 INTERFACE ${CMAKE_DL_LIBS})
+endif()
+
+target_link_libraries(kp_nvtx_connector CUDA::nvtx3)
--- a/.github/workflows/build-with-kokkos.yml
+++ b/.github/workflows/build-with-kokkos.yml
@@ -16,7 +16,7 @@ jobs:
           - image: nvidia/cuda:12.1.0-devel-ubuntu22.04
             preset: Cuda
             compiler: default
-          - image: nvidia/cuda:12.2.0-devel-ubuntu22.04
+          - image: nvidia/cuda:12.9.0-devel-ubuntu22.04
             preset: Cuda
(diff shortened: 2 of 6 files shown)


masterleinad:
Testing on `godbolt` suggests the nvtx3 header is available in CUDA 11.0.2.

masterleinad:
Looks like that `CMake` target was introduced in 3.25.

maartenarnst:
> Looks like that `CMake` target was introduced in 3.25.

Indeed, and the `CMake` release notes for that version say that `nvtx3` was available since `Cuda` 10.0.

Not sure how to deal with the issue of the `CMake` version. Should I write a fallback that creates the imported target manually?

profiling/nvtx-connector/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
 find_package(CUDAToolkit REQUIRED)
 kp_add_library(kp_nvtx_connector kp_nvtx_connector.cpp)
 
-target_link_libraries(kp_nvtx_connector CUDA::nvToolsExt)
+target_link_libraries(kp_nvtx_connector CUDA::nvtx3)
```
masterleinad, line 11: We could try something like
```suggestion
if (TARGET CUDA::nvtx3)
  target_link_libraries(kp_nvtx_connector CUDA::nvtx3)
else()
  target_link_libraries(kp_nvtx_connector CUDA::nvToolsExt)
endif()
```
and try to see if the `nvToolsExt` target still exists with CUDA 12.9 and if so if it provides the header used below.

maartenarnst: Thanks for the code suggestion! However, unfortunately, the target `CUDA::nvToolsExt` is not defined as of Cuda 12.9. I've just pushed a version with a manually created imported target. Let's see.

maartenarnst:
@masterleinad, it looks like the workaround with the manually created imported target worked.

masterleinad (approved):
Fine with me. I don't think we would need to be very concerned that we are messing with the `CUDA` namespace in `CMake`.

maartenarnst:
OK, thanks a lot! Hi @dalg24, would you have a moment to take a look?

romintomasetti (dismissed):
Should we move the logic to a centralized place?

vlkale:
@maartenarnst @romintomasetti  Thanks for this! 

This fix is good for the foreseeable future. The nvtx-connector and/or CMake should just be updated if there is, say, an nvtx4 in the far future.

JBludau (approved):
Not a terrible fan of the repetition but it can be restructured later.

dalg24:
I am fine with this PR.  Romin, did you mean to request changes?

romintomasetti:
> Not a terrible fan of the repetition but it can be restructured later.

I was thinking along the same lines as @JBludau about avoiding repetition.

romintomasetti (commented):
But not blocking.

PR #292, opened 2025-06-02
Remove option to reuse the kokkos compiler

### LINKED ISSUE #274: Option `KokkosTools_REUSE_KOKKOS_COMPILER` does not work

JBludau:
Fixes #274 
Resetting the CXX compiler is not a good idea and goes against the idea of CMake.

~https://github.com/kokkos/kokkos/pull/8138 proposes a abi/flag compatibility check in Kokkos itself, the user would get a warning when using incompatible compilers from core. With this in place users will get a hint and~ the agreement in #274 is to remove the option, which is done here

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,10 +47,6 @@ option(KokkosTools_ENABLE_APEX    "Enable building Apex library"    OFF)
 option(KokkosTools_ENABLE_EXAMPLES "Build examples"                 OFF)
 option(KokkosTools_ENABLE_TESTS    "Build tests"                    OFF)
 
-# Advanced settings
-option(KokkosTools_REUSE_KOKKOS_COMPILER "Set the compiler and flags based on installed Kokkos settings" OFF)
-mark_as_advanced(KokkosTools_REUSE_KOKKOS_COMPILER)
-
 # Fetch Kokkos options:
 acquire_kokkos_config()
 if(DEFINED Kokkos_FOUND_MSG)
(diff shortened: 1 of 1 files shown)


masterleinad (approved):

JBludau:
Since https://github.com/kokkos/kokkos/pull/8138 is closed, this is just a removal of the option

PR #293, opened 2025-06-11
Update Makefiles to nvtx3

ethan-puyaubreau:
Fixes Makefiles for CUDA tools according to updates made in PR #290 .

### DIFF
--- a/profiling/nvtx-connector/Makefile
+++ b/profiling/nvtx-connector/Makefile
@@ -1,7 +1,6 @@
 CXX=g++
 CXXFLAGS=-O3 -std=c++11 -g -I$(CUDA_ROOT)/include
 LDFLAGS=-L$(CUDA_ROOT)/lib64
-LIBS=-lnvtx3interop
 SHARED_CXXFLAGS=-shared -fPIC
 
 all: kp_nvtx_connector.so
@@ -12,7 +11,7 @@ CXXFLAGS+=-I${MAKEFILE_PATH} -I${MAKEFILE_PATH}../../common/makefile-only -I${MA
 
 kp_nvtx_connector.so: ${MAKEFILE_PATH}kp_nvtx_connector.cpp
 	$(CXX) $(SHARED_CXXFLAGS) $(CXXFLAGS) $(LDFLAGS) \
-		-o $@ ${MAKEFILE_PATH}kp_nvtx_connector.cpp $(LIBS)
+		-o $@ ${MAKEFILE_PATH}kp_nvtx_connector.cpp
 
 clean:
 	rm -f kp_nvtx_connector.so
--- a/profiling/nvtx-connector/Makefile
+++ b/profiling/nvtx-connector/Makefile
@@ -1,7 +1,7 @@
 CXX=g++
 CXXFLAGS=-O3 -std=c++11 -g -I$(CUDA_ROOT)/include
 LDFLAGS=-L$(CUDA_ROOT)/lib64
-LIBS=-lnvToolsExt
+LIBS=-lnvtx3interop
 SHARED_CXXFLAGS=-shared -fPIC
 
 all: kp_nvtx_connector.so
--- a/profiling/nvtx-focused-connector/Makefile
+++ b/profiling/nvtx-focused-connector/Makefile
@@ -1,7 +1,7 @@
 CXX=g++
 CXXFLAGS=-O3 -std=c++11 -g -I$(CUDA_ROOT)/include -I./
 LDFLAGS=-L$(CUDA_ROOT)/lib64
-LIBS=-lnvtx3interop -lcudart
+LIBS=-lcudart
 SHARED_CXXFLAGS=-shared -fPIC
 
 all: kp_nvtx_focused_connector.so
--- a/profiling/nvtx-focused-connector/Makefile
+++ b/profiling/nvtx-focused-connector/Makefile
@@ -1,7 +1,7 @@
 CXX=g++
 CXXFLAGS=-O3 -std=c++11 -g -I$(CUDA_ROOT)/include -I./
 LDFLAGS=-L$(CUDA_ROOT)/lib64
-LIBS=-lnvToolsExt -lcudart
+LIBS=-lnvtx3interop -lcudart
 SHARED_CXXFLAGS=-shared -fPIC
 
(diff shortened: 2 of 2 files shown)


ethan-puyaubreau:
Side note: The official NVTX documentation (https://nvidia.github.io/NVTX/) clearly states that NVTX3 is now a **header-only library** for C and C++.

While I've kept `-lnvtx3interop` for now to ensure immediate compatibility, it's very likely that removing the linking to this library entirely would be possible in the future, as NVTX3 no longer requires binary linking. This would simplify the Makefile.

JBludau:
The doc sounds to me that nvtx3 already is header only. Can you check if adding the headers is already enough to make it work? #290 seems to be using the header only one

ethan-puyaubreau:
Tried on my latest commit, seems to be working indeed.

ethan-puyaubreau:
It seems to be working without -lcudart too, but I'm not sure about this one being header only.

JBludau:
> It seems to be working without -lcudart too, but I'm not sure about this one being header only.

Did you check that with a compiler that does not automatically add the cuda runtime?

ethan-puyaubreau:
Not yet, I'll check while the testing workflow tries to see if there's any issue on its side too.

ethan-puyaubreau:
Ok, just checked on my system, it doesn't work. I did find that this library isn't header only at all too, so I'll revert this commit to only focus on nvtx3.

JBludau (approved):

masterleinad:
We could also think about retiring the Makefiles in this repository as well.

JBludau:
@vlkale are there plans to retire the Makefiles for the tools?

vlkale:
> @vlkale are there plans to retire the Makefiles for the tools?

The Makefiles have served Kokkos Tools well as an alternative to CMake IMO, including for this nvtx-connector. These are files from David, and this was actually the only way to use Kokkos Tools until a couple years ago.

However, with CMake build system in place: I understand the need to maintain these Makefiles for vendor connectors like this one.

@ethan-puyaubreau I agree with you keeping the binary linking in the Makefiles for now. 

 I have a mild preference to keeping at least some of the Makefiles alongside the CMake build system such as the one in kernel-logger.

Broadly, I suggest that if we do want to remove the Makefiles entirely for all tooling libraries and connectors of Kokkos Tools, or even just for nvtx-connector: we put that removal in a separate PR, as it would be a fairly major change. @masterleinad @ethan-puyaubreau let me know if that is an OK path forward.

ethan-puyaubreau:
Hello, that might be a good idea indeed/ @masterleinad if that's OK for you that's OK for me.

dalg24 (approved):

masterleinad:
> Hello, that might be a good idea indeed/ @masterleinad if that's OK for you that's OK for me.

The biggest problem with retiring/deprecating anything here is that we still don't have any release schedule for this repository.

vlkale:
> > Hello, that might be a good idea indeed/ @masterleinad if that's OK for you that's OK for me.
> 
> The biggest problem with retiring/deprecating anything here is that we still don't have any release schedule for this repository.

@masterleinad The consensus I gathered from previous discussions with the Kokkos team (including @crtrott and @dalg24 - as well as you, I think?) over the last couple years is that a release schedule for Kokkos Tools isn't needed and doesn't make sense generally. 

The reasoning is that this Kokkos Tools repo is a collection of in-house (Kokkos-team-developed) tool libraries like SpaceTimeStack _and_ **community-driven** tool libraries, e.g., the Kokkos Tools Variorium connector is not developed by the Kokkos team directly. We can't force a Kokkos Tools release cycle on these external vendors contributing Kokkos Tools connectors. Instead of release cycles, there have been roughly bi-annual presentations with updates to new connectors and enhancements at venues like Kokkos User Group, SC, etc. 

I also think that more maturity and testing infrastructure is important to do before setting up a release cycle. If there is a release cycle, maybe we be in 'version 0' until there is enough testing and infrastructure to establish the maturity of Kokkos Tools?

Do you think probing the Kokkos (and Kokkos Tools) team again how it feels about release cycles would be good to do? I'd be happy to help in seeing what the best path forward is. 

@ethan-puyaubreau @JBludau

masterleinad:
> The reasoning is that this Kokkos Tools repo is a collection of in-house (Kokkos-team-developed) tool libraries like SpaceTimeStack _and_ **community-driven** tool libraries, e.g., the Kokkos Tools Variorium connector is not developed by the Kokkos team directly. We can't force a Kokkos Tools release cycle on these external vendors contributing Kokkos Tools connectors. Instead of release cycles, there have been roughly bi-annual presentations with updates to new connectors and enhancements at venues like Kokkos User Group, SC, etc.

The point isn't to setup a plan for contributing particular features to Kokkos Tools but rather to indicate which commits/versions work with particular Kokkos releases. Moving files between Kokkos Core and Kokkos Tools resulted incompatibilities for example. Also, having releases allows us to deprecate and remove functionality as we like again referencong particular commits/versions.

vlkale:
Oh, I see - yes, this is important 

> > The reasoning is that this Kokkos Tools repo is a collection of in-house (Kokkos-team-developed) tool libraries like SpaceTimeStack _and_ **community-driven** tool libraries, e.g., the Kokkos Tools Variorium connector is not developed by the Kokkos team directly. We can't force a Kokkos Tools release cycle on these external vendors contributing Kokkos Tools connectors. Instead of release cycles, there have been roughly bi-annual presentations with updates to new connectors and enhancements at venues like Kokkos User Group, SC, etc.
> 
> The point isn't to setup a plan for contributing particular features to Kokkos Tools but rather to indicate which commits/versions work with particular Kokkos releases. Moving files between Kokkos Core and Kokkos Tools resulted incompatibilities for example. Also, having releases allows us to deprecate and remove functionality as we like again referencong particular commits/versions.

Oh, I see - yes, I agree and this makes sense. If that's the solution we established given:

https://github.com/kokkos/kokkos-tools/issues/275#issuecomment-2422634745 

then I am with you on this. Yes, we ought to consider some sort of Kokkos Tools release cycle to handle which git commit hash of Kokkos Tools works with a particular Kokkos release. I also see your point about deprecating functionality, but handling the above linked problem is probably the bigger thing in my mind.

PR #298, opened 2025-07-14
Makes Windows build fail earlier with more descriptive error message

OlekRaymond:
Building on windows is not currently supported but this could be communicated better.

I have made a change that makes the windows build fail immediately with an appropriate error message in the CMake. All other checks for Windows have been left alone so that Windows support can be added at a later time.

Ticket came as an improvement to pull request #289 which has been rendered obsolete

Thank you

### DIFF
--- a/common/kernel-filter/kp_kernel_filter.cpp
+++ b/common/kernel-filter/kp_kernel_filter.cpp
@@ -290,7 +290,7 @@ extern "C" void kokkosp_begin_parallel_reduce(const char* name,
                                               uint64_t* kID) {
   if (filterKernels) {
     if (kokkospFilterMatch(name)) {
-      if (NULL != beginScanCallee) {
+      if (NULL != beginReduceCallee) {
         (*beginReduceCallee)(name, devID, kID);
         activeKernels.insert(*kID);
       } else {
@@ -300,7 +300,7 @@ extern "C" void kokkosp_begin_parallel_reduce(const char* name,
       *kID = nextKernelID++;
     }
   } else {
-    if (NULL != beginScanCallee) {
+    if (NULL != beginReduceCallee) {
       (*beginReduceCallee)(name, devID, kID);
       activeKernels.insert(*kID);
     } else {


.gitignore
```diff
@@ -1,3 +1,5 @@
 *~
 *.so
 /src/tools/simple-kernel-timer/kp_reader
+# Expected build dir
+/build
```
JBludau: why expect this? We should not do that

CMakeLists.txt
```diff
@@ -2,6 +2,11 @@ cmake_minimum_required(VERSION 3.16 FATAL_ERROR)
 
 project(KokkosTools CXX)
 
+# Users cannot use Kokkos tools on windows, we want to test things on windows/msvc however
+if(WIN32 AND NOT DEFINED KOKKOS_TOOLS_TEST_WINDOWS)
+  message(FATAL_ERROR "Kokkos tools is not supported on Windows, please attempt to profile on expected hardware")
+endif()
+
```
JBludau: Does it even build on windows?

.gitignore
OlekRaymond: It's a very common name for a CMake build dir (I have yet to see anything else) and the suggested name in the documentation.

Adding it to the gitignore reduces the chance of it getting checked in, even if it is deleted soon after, unless the commit is squashed there is that artifact left around in the repo forever which can slow things down.

I can remove it though.

CMakeLists.txt
OlekRaymond: Yes,

The libs all build to completion, the `.so` objects are skipped and the build can pass successfully (all require certain cmake options)

At the moment you'd have to read the documentation to figure out that it's not supported*  , which is fine but lots of people don't so it's easier to just 

\* The user has to find out how to use it, find out that requires `.so`s then find out the `so`s don't build on windows (through a cmake warning) so this just makes it more clear

.gitignore
JBludau: There is a lot of comments on how to use CMake on Stackoverflow that are ... not super well researched.
CMake has `cmake -B <builddir>` and people can set to to whatever they want. My IDE puts part of the build config into the name of the folder. 
When something is in the `gitignore`, it will also no longer show up in `git status`. And if someone commits his build directory and creates a pullrequest here we will see it in the diff - I see little chance of it ending in the repo

CMakeLists.txt
JBludau: Maybe I don't understand ... but isn't `.so` the ending linux uses for shared objects? for windows it should be `.dll`, right? Can the resulting shared object be used on windows or are they broken?

.gitignore
OlekRaymond: > I see little chance of it ending in the repo

I have seen it happen before but you clearly have faith in programmers so too I will be optimistic and remove it!

FYI, people do often also add their IDE dirs (especially visual studio) to the gitignore, adding both build and IDE stuff has not caused problems for me in the past

CMakeLists.txt
OlekRaymond: >  .so the ending Linux uses for shared objects? for windows it should be .dll,

Yes they are the same concept and solve the same problems but the syntax and lookup changes between the two.

> Can the resulting shared object be used on window

No, In short we need some (minimal) code changes to make the `.dll` useful. I am considering doing this but I expect would see little user interaction because they should be profiling on expected hardware (like HPCs) which is rarely running Windows. 

Full disclosure: I work for a consultancy which has customers that like Windows, I'm not just a Linux hater :)

.gitignore
dalg24: I am generally against having a .gitignore in the repo.
This is something that developers should handle on their own, either via `.git/info/exclude` or via their own global git configuration.
Please drop that change.

CMakeLists.txt
JBludau: Yeah, I think the tools don't really have windows users on windows machines. Do you have seen them be used on WSL?
If the tools don't work on windows, maybe we should just have CMake error out

dalg24: Where would that variable be defined?
Why is it not sufficient to use `KokkosTools_ENABLE_TESTS={ON,OFF}`

OlekRaymond: > them be used on WSL

I remember reading (I don't know if it was Kokkos docs or other docs) not to profile in non-target environments so I think it's moot. I see no reason you can't profile a Kokkos executable built in WSL from WSL though.

> we should just have CMake error out

I agree, (with no intention to be rude) that is the point of the ticket. :)

OlekRaymond: > Where would that variable be defined?

I has to be given in the CMake invocation. i.e. `cmake -B something -S . -DKOKKOS_TOOLS_TEST_WINDOWS=ON`

> Why is it not sufficient to use KokkosTools_ENABLE_TESTS={ON,OFF}

I wanted something unique:
* so a user is unlikely to enable it by mistake and not see the error message.
* a dev is less likely to enable it
* since no devs are working on windows it won't be seen as excessive.
* it is unlikely to have any effect on a pipeline or future commits
* it does not clutter the help dialog in cmake like an option would as it is not intended for users

.gitignore
OlekRaymond: It has been removed

CMakeLists.txt
masterleinad: I'd suggest to add the option in a pull request that would actually add some `Windows` support.

OlekRaymond: I did think about that but it makes all the other checks for windows moot

I think it's fine here for now (does no harm and isn't particularly invasive)

janciesko:
```suggestion
if(WIN32)
  message(FATAL_ERROR "Kokkos Tools are not supported on Windows")
endif()
```

janciesko (changes requested):

CMakeLists.txt
janciesko: @masterleinad Let's just error out for now?

PR #299, opened 2025-08-15
Energy profiling tools: Core infrastructure with timing tool and export capabilities

ethan-puyaubreau:
This PR introduces the foundational infrastructure for energy profiling tools in Kokkos:

Features:
- Timing Infrastructure: Timing system with `TimingInfo` structure and region tracking
- State Management: Thread-safe `EnergyProfilerState` singleton for managing active/completed timing regions
- Region Types: Support for ParallelFor, ParallelReduce, ParallelScan, DeepCopy, and UserRegion profiling
- CSV Export: `timing_export` module for exporting timing data to CSV files with summary statistics

Implementation:
- Main Library: `kp_energy_profiler` - Core profiling functionality
- Utilities: `timing_utils.hpp/cpp` - State management and helper functions  
- Export: `timing_export.hpp/cpp` - CSV export and summary generation
- No External Dependencies: Self-contained implementation using only standard C++ libraries

Architecture:
This infrastructure provides a clean separation between timing collection, state management, and data export, making it easy for future energy monitoring providers to integrate their functionality. The design supports both synchronous and asynchronous profiling patterns (if combined with #300).

### DIFF
--- a/profiling/energy-profiler/CMakeLists.txt
+++ b/profiling/energy-profiler/CMakeLists.txt
@@ -1,0 +1,5 @@
+kp_add_library(kp_energy_profiler
+    kp_energy_profiler.cpp
+    timing_utils.cpp
+    timing_export.cpp
+)
--- a/profiling/energy-profiler/kp_energy_profiler.cpp
+++ b/profiling/energy-profiler/kp_energy_profiler.cpp
@@ -58,7 +58,8 @@ void start_region(const std::string& name, RegionType type, uint64_t id) {
 
 // End last region of given type
 void end_region_by_type(RegionType type_to_end) {
-  auto& state = EnergyProfilerState::get_instance();
+  auto end_time = std::chrono::high_resolution_clock::now();
+  auto& state   = EnergyProfilerState::get_instance();
   std::lock_guard<std::mutex> lock(state.get_mutex());
   auto& active_regions = state.get_active_regions();
   if (active_regions.empty()) return;
@@ -69,7 +70,7 @@ void end_region_by_type(RegionType type_to_end) {
   if (it != active_regions.rend()) {
     auto region = *it;
     active_regions.erase(std::next(it).base());
-    region.end_time = std::chrono::high_resolution_clock::now();
+    region.end_time = end_time;
     state.get_completed_timings().push_back(region);
   }
 }
--- a/profiling/energy-profiler/kp_energy_profiler.cpp
+++ b/profiling/energy-profiler/kp_energy_profiler.cpp
@@ -46,82 +46,65 @@ void log_verbose(const std::string& message) {
 
 // Start a region
 void start_region(const std::string& name, RegionType type, uint64_t id) {
-  try {
-    TimingInfo region;
-    region.name       = name;
-    region.type       = type;
-    region.start_time = std::chrono::high_resolution_clock::now();
-    region.id         = id;
-    auto& state       = EnergyProfilerState::get_instance();
-    std::lock_guard<std::mutex> lock(state.get_mutex());
-    state.get_active_regions().push_back(region);
-  } catch (const std::exception& e) {
-    std::cerr << "Error in start_region: " << e.what() << std::endl;
-  }
+  TimingInfo region;
+  region.name       = name;
+  region.type       = type;
+  region.start_time = std::chrono::high_resolution_clock::now();
+  region.id         = id;
+  auto& state       = EnergyProfilerState::get_instance();
+  std::lock_guard<std::mutex> lock(state.get_mutex());
+  state.get_active_regions().push_back(region);
 }
 
 // End last region of given type
 void end_region_by_type(RegionType type_to_end) {
-  try {
-    auto& state = EnergyProfilerState::get_instance();
-    std::lock_guard<std::mutex> lock(state.get_mutex());
-    auto& active_regions = state.get_active_regions();
-    if (active_regions.empty()) return;
-    auto it = std::find_if(active_regions.rbegin(), active_regions.rend(),
-                           [type_to_end](const TimingInfo& region) {
-                             return region.type == type_to_end;
-                           });
-    if (it != active_regions.rend()) {
-      auto region = *it;
-      active_regions.erase(std::next(it).base());
-      region.end_time = std::chrono::high_resolution_clock::now();
-      state.get_completed_timings().push_back(region);
-    }
-  } catch (const std::exception& e) {
-    std::cerr << "Error in end_region_by_type: " << e.what() << std::endl;
+  auto& state = EnergyProfilerState::get_instance();
+  std::lock_guard<std::mutex> lock(state.get_mutex());
+  auto& active_regions = state.get_active_regions();
+  if (active_regions.empty()) return;
+  auto it = std::find_if(active_regions.rbegin(), active_regions.rend(),
+                         [type_to_end](const TimingInfo& region) {
+                           return region.type == type_to_end;
+                         });
+  if (it != active_regions.rend()) {
+    auto region = *it;
+    active_regions.erase(std::next(it).base());
+    region.end_time = std::chrono::high_resolution_clock::now();
+    state.get_completed_timings().push_back(region);
   }
 }
 
 // End region by id
 void end_region_with_id(uint64_t expected_id) {
-  try {
-    auto end_time = std::chrono::high_resolution_clock::now();
-    auto& state   = EnergyProfilerState::get_instance();
-    std::lock_guard<std::mutex> lock(state.get_mutex());
-    auto& active_regions = state.get_active_regions();
-    auto it = std::find_if(active_regions.begin(), active_regions.end(),
-                           [expected_id](const TimingInfo& region) {
-                             return region.id == expected_id;
-                           });
-    if (it != active_regions.end()) {
-      auto region     = *it;
-      region.end_time = end_time;
-      active_regions.erase(it);
-      state.get_completed_timings().push_back(region);
-    } else {
-      std::cerr << "Warning: No active region found with ID " << expected_id
-                << "\n";
-    }
-  } catch (const std::exception& e) {
-    std::cerr << "Error in end_region_with_id: " << e.what() << std::endl;
+  auto end_time = std::chrono::high_resolution_clock::now();
+  auto& state   = EnergyProfilerState::get_instance();
+  std::lock_guard<std::mutex> lock(state.get_mutex());
+  auto& active_regions = state.get_active_regions();
+  auto it = std::find_if(active_regions.begin(), active_regions.end(),
+                         [expected_id](const TimingInfo& region) {
+                           return region.id == expected_id;
+                         });
+  if (it != active_regions.end()) {
+    auto region     = *it;
+    region.end_time = end_time;
+    active_regions.erase(it);
+    state.get_completed_timings().push_back(region);
+  } else {
+    std::cerr << "Warning: No active region found with ID " << expected_id
+              << "\n";
   }
 }
 
 // Get all completed timings
 std::vector<TimingInfo> get_all_timings() {
-  try {
-    auto& state = EnergyProfilerState::get_instance();
-    std::lock_guard<std::mutex> lock(state.get_mutex());
-    std::vector<TimingInfo> all_timings = state.get_completed_timings();
-    std::sort(all_timings.begin(), all_timings.end(),
-              [](const TimingInfo& a, const TimingInfo& b) {
-                return a.start_time < b.start_time;
-              });
-    return all_timings;
-  } catch (const std::exception& e) {
-    std::cerr << "Error in get_all_timings: " << e.what() << std::endl;
-    return {};
-  }
+  auto& state = EnergyProfilerState::get_instance();
+  std::lock_guard<std::mutex> lock(state.get_mutex());
+  std::vector<TimingInfo> all_timings = state.get_completed_timings();
+  std::sort(all_timings.begin(), all_timings.end(),
+            [](const TimingInfo& a, const TimingInfo& b) {
+              return a.start_time < b.start_time;
+            });
+  return all_timings;
 }
 
 }  // namespace EnergyProfiler
--- a/profiling/energy-profiler/kp_energy_profiler.cpp
+++ b/profiling/energy-profiler/kp_energy_profiler.cpp
@@ -20,7 +20,6 @@
 #include <cstdint>
 #include <cstdio>
 #include <cstdlib>
-#include <cstring>
 #include <iostream>
 #include <mutex>
 #include <string>
@@ -93,7 +92,8 @@ void end_region_by_type(RegionType type_to_end) {
 // End region by id
 void end_region_with_id(uint64_t expected_id) {
   try {
-    auto& state = EnergyProfilerState::get_instance();
+    auto end_time = std::chrono::high_resolution_clock::now();
+    auto& state   = EnergyProfilerState::get_instance();
     std::lock_guard<std::mutex> lock(state.get_mutex());
     auto& active_regions = state.get_active_regions();
     if (active_regions.empty()) {
@@ -106,9 +106,9 @@ void end_region_with_id(uint64_t expected_id) {
                              return region.id == expected_id;
                            });
     if (it != active_regions.end()) {
-      auto region = *it;
+      auto region     = *it;
+      region.end_time = end_time;
       active_regions.erase(it);
-      region.end_time = std::chrono::high_resolution_clock::now();
       state.get_completed_timings().push_back(region);
     } else {
       std::cerr << "Warning: No active region found with ID " << expected_id
@@ -154,7 +154,9 @@ void kokkosp_init_library(const int loadSeq, const uint64_t interfaceVer,
                           Kokkos_Profiling_KokkosPDeviceInfo* deviceInfo) {
   (void)devInfoCount;
   (void)deviceInfo;
-  if (std::getenv("KOKKOS_TOOLS_ENERGY_VERBOSE")) {
+  const char* verbose_env = std::getenv("KOKKOS_TOOLS_ENERGY_VERBOSE");
+  if (verbose_env &&
+      (std::string(verbose_env) == "1" || std::string(verbose_env) == "ON")) {
     KokkosTools::EnergyProfiler::EnergyProfilerState::get_instance()
         .set_verbose_enabled(true);
   }
@@ -170,7 +172,8 @@ void kokkosp_finalize_library() {
   printf("Kokkos Energy Profiler: Finalizing library\n");
   std::string prefix = KokkosTools::EnergyProfiler::generate_prefix();
   auto all_timings   = KokkosTools::EnergyProfiler::get_all_timings();
-  KokkosTools::EnergyProfiler::print_all_timings_summary(all_timings);
+  KokkosTools::EnergyProfiler::print_all_timings_summary(
+      std::cout, all_timings.begin(), all_timings.end());
   KokkosTools::EnergyProfiler::export_all_timings_csv(
       all_timings, prefix + "_timing_data.csv");
   printf("Kokkos Energy Profiler: Library finalized\n");
--- a/profiling/energy-profiler/kp_energy_profiler.cpp
+++ b/profiling/energy-profiler/kp_energy_profiler.cpp
@@ -89,11 +89,6 @@ void end_region_with_id(uint64_t expected_id) {
     auto& state   = EnergyProfilerState::get_instance();
     std::lock_guard<std::mutex> lock(state.get_mutex());
     auto& active_regions = state.get_active_regions();
-    if (active_regions.empty()) {
-      std::cerr << "Warning: Attempting to end region with ID " << expected_id
-                << " but no active regions found.\n";
-      return;
-    }
     auto it = std::find_if(active_regions.begin(), active_regions.end(),
                            [expected_id](const TimingInfo& region) {
                              return region.id == expected_id;
--- a/profiling/energy-profiler/kp_energy_profiler.cpp
+++ b/profiling/energy-profiler/kp_energy_profiler.cpp
@@ -16,10 +16,6 @@
 
 #include <algorithm>
 #include <chrono>
-#include <cstdarg>
-#include <cstdint>
-#include <cstdio>
-#include <cstdlib>
 #include <iostream>
 #include <mutex>
 #include <string>
@@ -42,12 +38,9 @@ uint64_t generate_new_region_id() {
 }
 
 // Helper function for verbose logging
-void log_verbose(const char* format, ...) {
+void log_verbose(const std::string& message) {
   if (EnergyProfilerState::get_instance().get_verbose_enabled()) {
-    va_list args;
-    va_start(args, format);
-    vprintf(format, args);
-    va_end(args);
+    std::cout << message << std::endl;
   }
 }
 
@@ -160,23 +153,22 @@ void kokkosp_init_library(const int loadSeq, const uint64_t interfaceVer,
     KokkosTools::EnergyProfiler::EnergyProfilerState::get_instance()
         .set_verbose_enabled(true);
   }
-  printf(
-      "Kokkos Energy Profiler: Initializing with load sequence %d and "
-      "interface version %lu\n",
-      loadSeq, interfaceVer);
-  printf("Kokkos Energy Profiler: Library initialized\n");
+  std::cout << "Kokkos Energy Profiler: Initializing with load sequence "
+            << loadSeq << " and interface version " << interfaceVer
+            << std::endl;
+  std::cout << "Kokkos Energy Profiler: Library initialized" << std::endl;
 }
 
 // Library finalize
 void kokkosp_finalize_library() {
-  printf("Kokkos Energy Profiler: Finalizing library\n");
+  std::cout << "Kokkos Energy Profiler: Finalizing library" << std::endl;
   std::string prefix = KokkosTools::EnergyProfiler::generate_prefix();
   auto all_timings   = KokkosTools::EnergyProfiler::get_all_timings();
   KokkosTools::EnergyProfiler::print_all_timings_summary(
       std::cout, all_timings.begin(), all_timings.end());
   KokkosTools::EnergyProfiler::export_all_timings_csv(
       all_timings, prefix + "_timing_data.csv");
-  printf("Kokkos Energy Profiler: Library finalized\n");
+  std::cout << "Kokkos Energy Profiler: Library finalized" << std::endl;
 }
 
 // Begin parallel_for
@@ -192,9 +184,9 @@ void kokkosp_begin_parallel_for(const char* name, const uint32_t devID,
   KokkosTools::EnergyProfiler::start_region(
       name, KokkosTools::EnergyProfiler::RegionType::ParallelFor, *kID);
   KokkosTools::EnergyProfiler::log_verbose(
-      "Kokkos Energy Profiler: Started parallel_for '%s' on device %u with ID "
-      "%lu\n",
-      name, devID, *kID);
+      std::string("Kokkos Energy Profiler: Started parallel_for '") + name +
+      "' on device " + std::to_string(devID) + " with ID " +
+      std::to_string(*kID));
 }
 
 // End parallel_for
@@ -205,7 +197,8 @@ void kokkosp_end_parallel_for(const uint64_t kID) {
   }
   KokkosTools::EnergyProfiler::end_region_with_id(kID);
   KokkosTools::EnergyProfiler::log_verbose(
-      "Kokkos Energy Profiler: Ended parallel_for with ID %lu\n", kID);
+      std::string("Kokkos Energy Profiler: Ended parallel_for with ID ") +
+      std::to_string(kID));
 }
 
 // Begin parallel_scan
@@ -221,9 +214,9 @@ void kokkosp_begin_parallel_scan(const char* name, const uint32_t devID,
   KokkosTools::EnergyProfiler::start_region(
       name, KokkosTools::EnergyProfiler::RegionType::ParallelScan, *kID);
   KokkosTools::EnergyProfiler::log_verbose(
-      "Kokkos Energy Profiler: Started parallel_scan '%s' on device %u with ID "
-      "%lu\n",
-      name, devID, *kID);
+      std::string("Kokkos Energy Profiler: Started parallel_scan '") + name +
+      "' on device " + std::to_string(devID) + " with ID " +
+      std::to_string(*kID));
 }
 
 // End parallel_scan
@@ -234,7 +227,8 @@ void kokkosp_end_parallel_scan(const uint64_t kID) {
   }
   KokkosTools::EnergyProfiler::end_region_with_id(kID);
   KokkosTools::EnergyProfiler::log_verbose(
-      "Kokkos Energy Profiler: Ended parallel_scan with ID %lu\n", kID);
+      std::string("Kokkos Energy Profiler: Ended parallel_scan with ID ") +
+      std::to_string(kID));
 }
 
 // Begin parallel_reduce
@@ -250,9 +244,9 @@ void kokkosp_begin_parallel_reduce(const char* name, const uint32_t devID,
   KokkosTools::EnergyProfiler::start_region(
       name, KokkosTools::EnergyProfiler::RegionType::ParallelReduce, *kID);
   KokkosTools::EnergyProfiler::log_verbose(
-      "Kokkos Energy Profiler: Started parallel_reduce '%s' on device %u with "
-      "ID %lu\n",
-      name, devID, *kID);
+      std::string("Kokkos Energy Profiler: Started parallel_reduce '") + name +
+      "' on device " + std::to_string(devID) + " with ID " +
+      std::to_string(*kID));
 }
 
 // End parallel_reduce
@@ -263,7 +257,8 @@ void kokkosp_end_parallel_reduce(const uint64_t kID) {
   }
   KokkosTools::EnergyProfiler::end_region_with_id(kID);
   KokkosTools::EnergyProfiler::log_verbose(
-      "Kokkos Energy Profiler: Ended parallel_reduce with ID %lu\n", kID);
+      std::string("Kokkos Energy Profiler: Ended parallel_reduce with ID ") +
+      std::to_string(kID));
 }
 
 // Push user region
@@ -276,7 +271,8 @@ void kokkosp_push_profile_region(char const* regionName) {
   KokkosTools::EnergyProfiler::start_region(
       regionName, KokkosTools::EnergyProfiler::RegionType::UserRegion, new_id);
   KokkosTools::EnergyProfiler::log_verbose(
-      "Kokkos Energy Profiler: Pushed profile region '%s'\n", regionName);
+      std::string("Kokkos Energy Profiler: Pushed profile region '") +
+      regionName + "'");
 }
 
 // Pop user region
@@ -284,7 +280,7 @@ void kokkosp_pop_profile_region() {
   KokkosTools::EnergyProfiler::end_region_by_type(
       KokkosTools::EnergyProfiler::RegionType::UserRegion);
   KokkosTools::EnergyProfiler::log_verbose(
-      "Kokkos Energy Profiler: Popped profile region\n");
+      "Kokkos Energy Profiler: Popped profile region");
 }
 
 // Begin deep copy
@@ -300,9 +296,9 @@ void kokkosp_begin_deep_copy(Kokkos::Tools::SpaceHandle, const char* dst_name,
   KokkosTools::EnergyProfiler::start_region(
       name, KokkosTools::EnergyProfiler::RegionType::DeepCopy, new_id);
   KokkosTools::EnergyProfiler::log_verbose(
-      "Kokkos Energy Profiler: Started deep copy from '%s' to '%s' (size: %lu "
-      "bytes)\n",
-      src_name, dst_name, size);
+      std::string("Kokkos Energy Profiler: Started deep copy from '") +
+      src_name + "' to '" + dst_name + "' (size: " + std::to_string(size) +
+      " bytes)");
 }
 
 // End deep copy
@@ -310,7 +306,7 @@ void kokkosp_end_deep_copy() {
   KokkosTools::EnergyProfiler::end_region_by_type(
       KokkosTools::EnergyProfiler::RegionType::DeepCopy);
   KokkosTools::EnergyProfiler::log_verbose(
-      "Kokkos Energy Profiler: Ended deep copy\n");
+      "Kokkos Energy Profiler: Ended deep copy");
 }
 
 }  // extern "C"
--- a/profiling/energy-profiler/kp_energy_profiler.cpp
+++ b/profiling/energy-profiler/kp_energy_profiler.cpp
@@ -47,13 +47,14 @@ void log_verbose(const std::string& message) {
 // Start a region
 void start_region(const std::string& name, RegionType type, uint64_t id) {
   TimingInfo region;
-  region.name       = name;
-  region.type       = type;
-  region.start_time = std::chrono::high_resolution_clock::now();
-  region.id         = id;
-  auto& state       = EnergyProfilerState::get_instance();
+  region.name = name;
+  region.type = type;
+  region.id   = id;
+  auto& state = EnergyProfilerState::get_instance();
   std::lock_guard<std::mutex> lock(state.get_mutex());
   state.get_active_regions().push_back(region);
+  state.get_active_regions().back().start_time =
+      std::chrono::high_resolution_clock::now();
 }
 
 // End last region of given type
--- a/profiling/energy-profiler/kp_energy_profiler.cpp
+++ b/profiling/energy-profiler/kp_energy_profiler.cpp
@@ -1,0 +1,313 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#include <algorithm>
+#include <chrono>
+#include <cstdarg>
+#include <cstdint>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+#include <iostream>
+#include <mutex>
+#include <string>
+#include <vector>
+
+#include "kp_core.hpp"
+#include "timing_utils.hpp"
+#include "timing_export.hpp"
+
+namespace KokkosTools {
+namespace EnergyProfiler {
+
+// Helper function to generate new region ID
+uint64_t generate_new_region_id() {
+  auto& state = EnergyProfilerState::get_instance();
+  std::lock_guard<std::mutex> lock(state.get_mutex());
+  uint64_t current_id = state.get_next_region_id();
+  state.increment_next_region_id();
+  return current_id;
+}
+
+// Helper function for verbose logging
+void log_verbose(const char* format, ...) {
+  if (EnergyProfilerState::get_instance().get_verbose_enabled()) {
+    va_list args;
+    va_start(args, format);
+    vprintf(format, args);
+    va_end(args);
+  }
+}
+
+// Start a region
+void start_region(const std::string& name, RegionType type, uint64_t id) {
+  try {
+    TimingInfo region;
+    region.name       = name;
+    region.type       = type;
+    region.start_time = std::chrono::high_resolution_clock::now();
+    region.id         = id;
+    auto& state       = EnergyProfilerState::get_instance();
+    std::lock_guard<std::mutex> lock(state.get_mutex());
+    state.get_active_regions().push_back(region);
+  } catch (const std::exception& e) {
+    std::cerr << "Error in start_region: " << e.what() << std::endl;
+  }
+}
+
+// End last region of given type
+void end_region_by_type(RegionType type_to_end) {
+  try {
+    auto& state = EnergyProfilerState::get_instance();
+    std::lock_guard<std::mutex> lock(state.get_mutex());
+    auto& active_regions = state.get_active_regions();
+    if (active_regions.empty()) return;
+    auto it = std::find_if(active_regions.rbegin(), active_regions.rend(),
+                           [type_to_end](const TimingInfo& region) {
+                             return region.type == type_to_end;
+                           });
+    if (it != active_regions.rend()) {
+      auto region = *it;
+      active_regions.erase(std::next(it).base());
+      region.end_time = std::chrono::high_resolution_clock::now();
+      state.get_completed_timings().push_back(region);
+    }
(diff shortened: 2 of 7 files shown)


profiling/energy-profiler/common/error_handling.hpp
```diff
@@ -0,0 +1,101 @@
...
+ */
+#define ENERGY_PROFILER_LOG_INFO(component, msg) \
+  KokkosTools::EnergyProfiler::Logger::info(component, msg)
+
+#define ENERGY_PROFILER_LOG_WARNING(component, msg) \
+  KokkosTools::EnergyProfiler::Logger::warning(component, msg)
+
+#define ENERGY_PROFILER_LOG_ERROR(component, msg) \
+  KokkosTools::EnergyProfiler::Logger::error(component, msg)
+
```
JBludau: ok, if it is only used via these macros this is ok. But they should be prefixed with KOKKOS_TOOLS or similar (consistent in the lib)

profiling/energy-profiler/common/filename_prefix.cpp
```diff
@@ -0,0 +1,8 @@
+#include "filename_prefix.hpp"
+
+std::string generate_prefix() {
+  char hostname[256];
+  gethostname(hostname, 256);
```
JBludau: does that work on all UNIX?

profiling/energy-profiler/common/timer_system.cpp
```diff
@@ -0,0 +1,318 @@
...
+//
+//@HEADER
+
+#include "timer_system.hpp"
+
+#include <cstdio>
+#include <iostream>
+
+namespace KokkosTools {
+namespace Timer {
```
JBludau: hmm `Timer` might be a bit of a general name and raise the question why this is repeated from the other timer

profiling/energy-profiler/common/timer_system.cpp
```diff
@@ -0,0 +1,318 @@
...
+
+// === CSV Export Functions ===
+
+void export_kernels_csv(const std::deque<TimingInfo>& timings,
```
JBludau: these are repeated and could be consolidated, right?

profiling/energy-profiler/common/timer_system.cpp
```diff
@@ -0,0 +1,318 @@
...
+
+    std::cout << "| " << format_table_cell(info.name, 38) << "| "
+              << format_table_cell(type, 16) << "| "
+              << format_table_cell(std::to_string(start_ms), 19) << "| "
+              << format_table_cell(std::to_string(end_ms), 19) << "| "
+              << format_table_cell(std::to_string(duration_ms), 13) << "|\n";
+  }
+}
+
+void print_regions_summary(const std::deque<TimingInfo>& regions) {
```
JBludau: same here

JBludau: and they could be put into a util file

profiling/energy-profiler/kokkos-tools/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+    ../common/filename_prefix.cpp
+)
+
+target_include_directories(kp_energy_kernel_timer PRIVATE
+    ${CMAKE_CURRENT_SOURCE_DIR}
+)
+
+# Add verbose output option
+if(ENABLE_ENERGY_PROFILER_VERBOSE)
+    target_compile_definitions(kp_energy_kernel_timer PRIVATE ENABLE_VERBOSE_OUTPUT)
```
JBludau: These should be prefixed with KOKKOS_TOOLS or something similar

profiling/energy-profiler/kokkos-tools/kp_energy_kernel_timer.cpp
```diff
@@ -0,0 +1,204 @@
...
+#include <iostream>
+
+#include "kp_core.hpp"
+#include "../common/timer_system.hpp"
+#include "../common/filename_prefix.hpp"
+
+namespace KokkosTools {
+namespace KernelTimer {
+
+// --- Core Initialization ---
```
JBludau: that is a misleading comment

JBludau (commented):
first pass

profiling/energy-profiler/common/timer_system.cpp
```diff
@@ -75,129 +73,104 @@ void export_regions_csv(const std::deque<TimingInfo>& timings,
...
-  for (const auto& info : regions) {
-    auto [start_ms, end_ms] = get_timing_ms(info);
-    auto duration_ms        = get_duration_ms(info);
-
-    std::cout << "| " << format_table_cell(info.name, 38) << "| "
-              << format_table_cell(std::to_string(start_ms), 19) << "| "
-              << format_table_cell(std::to_string(end_ms), 19) << "| "
-              << format_table_cell(std::to_string(duration_ms), 13) << "|\n";
-  }
+  print_timings_summary_generic(regions, "REGIONS", false);
```
JBludau: at that point the question is: do we even need the differentiation into multiple functions if they just forward

profiling/energy-profiler/kokkos-tools/kp_energy_kernel_timer.cpp
```diff
@@ -30,10 +29,10 @@
...
-// --- Core Initialization ---
-Timer::KernelTimerTool timer;
+// --- Core Timer Infrastructure ---
+EnergyProfiler::KernelTimerTool timer;
 
-#ifdef ENABLE_VERBOSE_OUTPUT
+#ifdef KOKKOS_TOOLS_ENABLE_VERBOSE_OUTPUT
 constexpr bool VERBOSE = true;
 #else
 constexpr bool VERBOSE = false;
```
JBludau: this can have a better name

profiling/energy-profiler/common/timer_system.cpp
ethan-puyaubreau: I don't really know on this one. Would we need to know the type of thing it is everytime and in this case we would have only one function that does it all, or something else? What do you think would be best?

profiling/energy-profiler/kokkos-tools/kp_energy_kernel_timer.cpp
ethan-puyaubreau: Something specific to the energy-profiling system I guess?

JBludau: yes. VERBOSE might even clash with some macro definitions

profiling/energy-profiler/common/timer_system.cpp
JBludau: if the type is also in the name of the call, we (at least as dev people) have to know the type anyways

profiling/energy-profiler/common/timer_system.cpp
```diff
@@ -0,0 +1,275 @@
...
+  auto end_ms = std::chrono::duration_cast<std::chrono::milliseconds>(
+                    info.end_time.time_since_epoch())
+                    .count();
+  return {start_ms, end_ms};
+}
+
+// Helper function to get duration in milliseconds
+long get_duration_ms(const TimingInfo& info) {
+  return info.duration.count() / 1000000;
+}
```
JBludau: isn't that converting ms to s?

profiling/energy-profiler/common/timer_system.cpp
```diff
@@ -0,0 +1,275 @@
...
+  active_regions_.push_back(region);
+}
+
+void KernelTimerTool::end_region() {
+  if (!active_regions_.empty()) {
+    auto region = active_regions_.back();
+    active_regions_.pop_back();
+    region.end_time = std::chrono::high_resolution_clock::now();
+    region.duration = std::chrono::duration_cast<std::chrono::nanoseconds>(
+        region.end_time - region.start_time);
```
JBludau: hmm this always ends the last region in the `dqueue`. should we not match it with the `name`?

ethan-puyaubreau:
@dalg24 @masterleinad Hello! I would need some review on this PR, being the baseline blocks needed the daemon system and energy measurement tools (that would also need review #300)

profiling/energy-profiler/common/error_handling.hpp
```diff
@@ -0,0 +1,101 @@
...
+  SUCCESS = 0,
+  PROVIDER_INIT_FAILED,
+  DEVICE_ACCESS_FAILED,
+  FILE_WRITE_FAILED,
+  INVALID_DEVICE_INDEX,
+  MEASUREMENT_FAILED,
+  UNKNOWN_ERROR
+};
+
+struct Result {
```
JBludau: Do we really need that? and if so, why not make it a pure aggregate? https://en.cppreference.com/w/cpp/language/aggregate_initialization.html

profiling/energy-profiler/common/error_handling.hpp
```diff
@@ -0,0 +1,101 @@
...
+  ErrorCode code;
+  std::string message;
+
+  Result() : code(ErrorCode::SUCCESS) {}
+  Result(ErrorCode c) : code(c) {}
```
JBludau: this one should be made `explicit` if the struct as a whole is even necessary

profiling/energy-profiler/common/error_handling.hpp
```diff
@@ -0,0 +1,101 @@
...
+  bool is_success() const { return code == ErrorCode::SUCCESS; }
+  bool is_error() const { return code != ErrorCode::SUCCESS; }
+
+  operator bool() const { return is_success(); }
+};
+
+/**
+ * @brief Simple logging utility
+ */
+class Logger {
```
JBludau: this one could be converted into a single free function

profiling/energy-profiler/common/error_handling.hpp
```diff
@@ -0,0 +1,101 @@
...
+ * @brief Helper macros for consistent error reporting
+ */
+#define KOKKOS_TOOLS_ENERGY_PROFILER_LOG_INFO(component, msg) \
+  KokkosTools::EnergyProfiler::Logger::info(component, msg)
+
+#define KOKKOS_TOOLS_ENERGY_PROFILER_LOG_WARNING(component, msg) \
+  KokkosTools::EnergyProfiler::Logger::warning(component, msg)
+
+#define KOKKOS_TOOLS_ENERGY_PROFILER_LOG_ERROR(component, msg) \
+  KokkosTools::EnergyProfiler::Logger::error(component, msg)
```
JBludau: why do you want to use them via macros and not directly?

profiling/energy-profiler/common/timer_system.cpp
```diff
@@ -0,0 +1,275 @@
...
+  if (file) {
+    fprintf(file, "%s\n", header.c_str());
+    for (const auto& timing : timings) {
+      auto start_ms = std::chrono::duration_cast<std::chrono::milliseconds>(
+                          timing.start_time.time_since_epoch())
+                          .count();
+      auto end_ms = std::chrono::duration_cast<std::chrono::milliseconds>(
+                        timing.end_time.time_since_epoch())
+                        .count();
+      auto duration_ms = timing.duration.count() / 1000000;
```
JBludau: same comment about converting from ms to s here

profiling/energy-profiler/common/timer_system.cpp
```diff
@@ -0,0 +1,275 @@
...
+  print_timings_summary(kernels, "KERNELS");
+}
+
+void print_regions_summary(const std::deque<TimingInfo>& regions) {
+  print_timings_summary(regions, "REGIONS");
+}
+
+void print_deepcopies_summary(const std::deque<TimingInfo>& deepcopies) {
+  print_timings_summary(deepcopies, "DEEP COPIES");
+}
```
JBludau: I am still not convinced that these are really necessary ... wherever they are called, the title has to be known by the user since it also is in the function name

profiling/energy-profiler/common/timer_system.cpp
```diff
@@ -0,0 +1,275 @@
...
+// === CSV Export Functions ===
+
+void export_kernels_csv(const std::deque<TimingInfo>& timings,
+                        const std::string& filename) {
+  export_timings_csv_generic(
+      timings, filename,
+      "name,type,start_time_epoch_ms,end_time_epoch_ms,duration_ms", true);
+  if (!timings.empty()) {
+    std::cout << "Timing data exported to " << filename << std::endl;
+  }
```
JBludau: maybe add an else with a print for clarity?

profiling/energy-profiler/common/timer_system.cpp
```diff
@@ -0,0 +1,275 @@
...
+    }
+    fclose(file);
+  } else {
+    std::cerr << "ERROR: Unable to open file " << filename << " for writing.\n";
+  }
+}
+
+// === CSV Export Functions ===
+
+void export_kernels_csv(const std::deque<TimingInfo>& timings,
```
JBludau: Same here, I am not sure if we really need different functions for this ... they all forward to the generic one and print a simple status ...

profiling/energy-profiler/kokkos-tools/kp_energy_kernel_timer.cpp
```diff
@@ -0,0 +1,205 @@
...
+// --- Core Timer Infrastructure ---
+EnergyProfiler::KernelTimerTool timer;
+
+#ifdef KOKKOS_TOOLS_ENABLE_VERBOSE_OUTPUT
+constexpr bool ENERGY_PROFILER_VERBOSE_OUTPUT = true;
+#else
+constexpr bool ENERGY_PROFILER_VERBOSE_OUTPUT = false;
+#endif
+
+std::string KOKKOS_PROFILE_LIBRARY_NAME =
```
JBludau: const?

profiling/energy-profiler/kokkos-tools/kp_energy_kernel_timer.cpp
```diff
@@ -0,0 +1,205 @@
...
+#include "../common/filename_prefix.hpp"
+
+namespace KokkosTools {
+namespace KernelTimer {
+
+// --- Core Timer Infrastructure ---
+EnergyProfiler::KernelTimerTool timer;
+
+#ifdef KOKKOS_TOOLS_ENABLE_VERBOSE_OUTPUT
+constexpr bool ENERGY_PROFILER_VERBOSE_OUTPUT = true;
```
JBludau: I am in general a fan of using all caps for macros and something like `EnergyProfiler::is_verbose()` for these functions

profiling/energy-profiler/kokkos-tools/kp_energy_kernel_timer.cpp
```diff
@@ -0,0 +1,205 @@
...
+  std::cout << "Kokkos Power Profiler: Library finalized\n";
+
+  std::string prefix = generate_prefix();
+
+  const auto& kernels = timer.get_kernel_timings();
+  KokkosTools::EnergyProfiler::print_kernels_summary(kernels);
+  KokkosTools::EnergyProfiler::export_kernels_csv(kernels,
+                                                  prefix + "_kernels.csv");
+
+  // Récapitulatif des régions
```
JBludau: :-D please use english as language for comments

ethan-puyaubreau: good find! I was not expecting that kind of typo there.

profiling/energy-profiler/common/timer_system.cpp
ethan-puyaubreau: no since the time is stored in nanoseconds in the first place

profiling/energy-profiler/common/timer_system.cpp
ethan-puyaubreau: same thing here, time is in nanoseconds in the first place

profiling/energy-profiler/common/timer_system.hpp
```diff
@@ -0,0 +1,110 @@
...
+  // Stack-based timing for robust region/kernel tracking
+  void start_region(const std::string& name, RegionType type, uint64_t id = 0);
+  void end_region();
+
+  // Getters for completed timings
+  const std::deque<TimingInfo>& get_kernel_timings() const;
+  const std::deque<TimingInfo>& get_region_timings() const;
+  const std::deque<TimingInfo>& get_deep_copy_timings() const;
+
+ private:
```
JBludau: Maybe add a comment about insertion times to show, that we chose `deque` for the constant insertion time independent of how many are inserted.

profiling/energy-profiler/kokkos-tools/kp_energy_kernel_timer.cpp
```diff
@@ -0,0 +1,211 @@
...
+  if (is_verbose_enabled()) {
+    std::cout << "Kokkos Power Profiler: Ended parallel_for with ID " << kID
+              << "\n";
+  }
+}
+
+void kokkosp_begin_parallel_scan(const char* name, const uint32_t devID,
+                                 uint64_t* kID) {
+  timer.begin_parallel_scan(name, devID, kID);
+  if (is_verbose_enabled()) {
```
JBludau: these could also be `constexpr`

profiling/energy-profiler/unit-tests/csv_export_test.cpp
```diff
@@ -0,0 +1,78 @@
...
+  timer.end_region();
+
+  // Test the CSV export functions
+  const auto& kernels    = timer.get_kernel_timings();
+  const auto& regions    = timer.get_region_timings();
+  const auto& deepcopies = timer.get_deep_copy_timings();
+
+  std::cout << "Found " << kernels.size() << " kernels" << std::endl;
+  std::cout << "Found " << regions.size() << " regions" << std::endl;
+  std::cout << "Found " << deepcopies.size() << " deep copies" << std::endl;
```
JBludau: in the tests we have gtest, see `/tests/CMakeLists.txt:113:target_link_libraries(test_common PUBLIC GTest::gtest GTest::gmock Kokkos::kokkos)`
maybe move the unit tests into the tests directory and then use something like `ASSERT_EQ` on the `size()` and print if it is not true (see https://github.com/kokkos/kokkos/blob/34d1af5cb9fbe11730aa641735bfa7ce9db8c6b9/core/unit_test/TestUniqueToken.hpp#L132)

profiling/energy-profiler/kokkos-tools/kp_energy_kernel_timer.cpp
ethan-puyaubreau: which part would be `constexpr`?

profiling/energy-profiler/common/timer_system.cpp
```diff
@@ -85,78 +122,37 @@ void export_timings_csv_generic(const std::deque<TimingInfo>& timings,
...
+    auto duration_ms        = get_duration_ms(timing_info);
+    std::string type        = region_type_to_string(timing_info.type);
 
-    std::cout << "| " << format_table_cell(info.name, 38) << "| "
+    std::cout << "| " << format_table_cell(timing_info.name, 38) << "| "
               << format_table_cell(type, 16) << "| "
               << format_table_cell(std::to_string(start_ms), 19) << "| "
               << format_table_cell(std::to_string(end_ms), 19) << "| "
               << format_table_cell(std::to_string(duration_ms), 13) << "|\n";
   }
```
JBludau: Hmm the width here is fixed ... although I like that it gives an easy for humans to read table, it would not be required for other programs to parse it.

profiling/energy-profiler/kokkos-tools/kp_energy_kernel_timer.cpp
JBludau: since `is_verbose_enabled() is a constexpr it should work in `if constexpr (...)`, see https://en.cppreference.com/w/cpp/language/if.html

JBludau (commented):
I think we are getting there

profiling/energy-profiler/common/error_handling.hpp
```diff
@@ -0,0 +1,46 @@
...
+                        const std::string& message) {
+  const char* level_str = "";
+  switch (level) {
+    case LogLevel::INFO: level_str = "INFO"; break;
+    case LogLevel::WARNING: level_str = "WARNING"; break;
+    case LogLevel::ERROR: level_str = "ERROR"; break;
+  }
+
+  std::cerr << "[" << level_str << "] " << component << ": " << message
+            << std::endl;
```
dalg24: Why do we print to the standard error output stream on INFO?

profiling/energy-profiler/common/filename_prefix.cpp
```diff
@@ -0,0 +1,8 @@
...
+  int pid = (int)getpid();
+  return std::string(hostname) + "-" + std::to_string(pid);
+}
```
dalg24: Fix your files' end

profiling/energy-profiler/common/filename_prefix.hpp
```diff
@@ -0,0 +1,6 @@
+#pragma once
+
+#include <string>
+#include <unistd.h>
+
+std::string generate_prefix();
```
dalg24: Use a namespace

profiling/energy-profiler/common/filename_prefix.hpp
```diff
@@ -0,0 +1,6 @@
+#pragma once
+
+#include <string>
+#include <unistd.h>
```
dalg24: These header includes are unnecessary/misplaced

profiling/energy-profiler/common/timer_system.hpp
```diff
@@ -0,0 +1,112 @@
...
+namespace EnergyProfiler {
+
+// Forward declarations
+enum class RegionType {
+  Unknown,
+  ParallelFor,
+  ParallelReduce,
+  ParallelScan,
+  DeepCopy,
+  UserRegion
```
dalg24: They are also user-defined sections but we could add them later.

profiling/energy-profiler/common/timer_system.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+// Data category for export and display
+enum class DataCategory { Kernels, Regions, DeepCopies };
+
+struct TimingInfo {
+  std::string name;
+  RegionType type;
+  std::chrono::high_resolution_clock::time_point start_time;
+  std::chrono::high_resolution_clock::time_point end_time;
+  std::chrono::nanoseconds duration;
```
dalg24: Isn't it a bit redundant?  Can't we deduce duration from the start and end time points?

profiling/energy-profiler/common/timer_system.hpp
```diff
@@ -0,0 +1,112 @@
...
+// Summary printing functions
+void print_timings_summary(const std::deque<TimingInfo>& timings,
+                           DataCategory category);
+
+// Unified Timer Tool Class
+class KernelTimerTool {
+ public:
+  KernelTimerTool()  = default;
+  ~KernelTimerTool() = default;
+
```
dalg24:
```suggestion
```

profiling/energy-profiler/common/timer_system.hpp
```diff
@@ -0,0 +1,112 @@
...
+  // Kokkos profiling interface implementation
+  void init_library(const int loadSeq, const uint64_t interfaceVer,
+                    const uint32_t devInfoCount,
```
dalg24: const-qualifiers are pointless on function declarations

profiling/energy-profiler/common/timer_system.hpp
```diff
@@ -0,0 +1,112 @@
...
+                    Kokkos_Profiling_KokkosPDeviceInfo* deviceInfo);
+  void finalize_library();
+
+  void begin_parallel_for(const char* name, const uint32_t devID, uint64_t kID);
```
dalg24: That does not look right.  Presumably you want to set the kernel ID but it is passed by value.

profiling/energy-profiler/kokkos-tools/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+    ../common/timer_system.cpp
+    ../common/filename_prefix.cpp
+)
+
+target_include_directories(kp_energy_kernel_timer PRIVATE
+    ${CMAKE_CURRENT_SOURCE_DIR}
+)
+
+# Add verbose output option
+if(KOKKOS_TOOLS_ENABLE_ENERGY_PROFILER_VERBOSE)
```
dalg24: Do we really want it to be a compile-time option or is it something we might want to be able to select at runtime?

profiling/energy-profiler/kokkos-tools/kp_energy_kernel_timer.cpp
```diff
@@ -0,0 +1,211 @@
...
+// --- Library Initialization/Finalization ---
+
+void kokkosp_init_library(const int loadSeq, const uint64_t interfaceVer,
+                          const uint32_t devInfoCount,
+                          Kokkos_Profiling_KokkosPDeviceInfo* deviceInfo) {
+  timer.init_library(loadSeq, interfaceVer, devInfoCount, deviceInfo);
+}
+
+void kokkosp_finalize_library() {
+  std::cout << "Kokkos Power Profiler: Finalizing library\n";
```
dalg24: "Energy" Profiler?

profiling/energy-profiler/kokkos-tools/kp_energy_kernel_timer.cpp
```diff
@@ -0,0 +1,211 @@
...
+EnergyProfiler::KernelTimerTool timer;
+
+#ifdef KOKKOS_TOOLS_ENABLE_VERBOSE_OUTPUT
+constexpr bool is_verbose_enabled() { return true; }
+#else
+constexpr bool is_verbose_enabled() { return false; }
+#endif
+
+const std::string KOKKOS_PROFILE_LIBRARY_NAME =
+    "Kokkos Kernel Timer for Energy Profiler";
```
dalg24: Never used, drop it.

profiling/energy-profiler/common/error_handling.hpp
dalg24, line 1: Did you mean to commit that file?
It does not seem like it is used anywhere.

profiling/energy-profiler/common/timer_system.cpp
```diff
@@ -0,0 +1,271 @@
...
+//
+//@HEADER
+
+#include "timer_system.hpp"
+
+#include <cstdio>
+#include <iostream>
+
+namespace KokkosTools {
+namespace EnergyProfiler {
```
dalg24: Fully qualify all function definitions and put everything else in an anonymous namespace.

profiling/energy-profiler/kokkos-tools/kp_energy_kernel_timer.cpp
```diff
@@ -0,0 +1,211 @@
...
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+/**
+ * Kokkos Energy Profiler - Kernel Timer
+ * Basic kernel timing tool for the energy profiling infrastructure
+ */
```
dalg24: Prefer C++ style comments (`// Bla`)

Or get rid of it.  It is not really saying anything particularly helpful.

profiling/energy-profiler/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+# Energy Profiler Infrastructure
+# This provides the foundational infrastructure for energy profiling tools:
+# - Timer system for kernel and region tracking
+# - Common utilities (filename generation, error handling)
+# - Tool interface definitions
+# - Basic kernel timer tool
+
+option(KOKKOS_TOOLS_ENABLE_ENERGY_PROFILER_VERBOSE "Enable verbose output for energy profiler tools" OFF)
```
dalg24: If we really want an option then we would spell it `KokkosTools_BLA`.

tests/energy-profiler/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
+# Energy Profiler Tests
+
+# Timer system test
+add_executable(test_energy_profiler_timer)
+target_sources(
+    test_energy_profiler_timer
+    PRIVATE
+        test_timer_system.cpp
+        ../../profiling/energy-profiler/common/timer_system.cpp
+        ../../profiling/energy-profiler/common/filename_prefix.cpp
```
dalg24: What does listing these as source do?
Why are we not just linking against `kp_energy_kernel_timer`?

profiling/energy-profiler/common/timer_system.cpp
```diff
@@ -0,0 +1,271 @@
...
+        std::string type = region_type_to_string(timing.type);
+        fprintf(file, "%s,%s,%ld,%ld,%ld\n", timing.name.c_str(), type.c_str(),
+                start_ms, end_ms, duration_ms);
+      } else {
+        fprintf(file, "%s,%ld,%ld,%ld\n", timing.name.c_str(), start_ms, end_ms,
+                duration_ms);
+      }
+    }
+    fclose(file);
+    std::cout << info.export_message << " " << filename << std::endl;
```
dalg24: Prefer `'\n'` to `std::endl` unless you mean to flush the stream

profiling/energy-profiler/common/timer_system.cpp
```diff
@@ -0,0 +1,271 @@
...
+                    info.end_time.time_since_epoch())
+                    .count();
+  return {start_ms, end_ms};
+}
+
+// Helper function to get duration in milliseconds
+// Note: duration is stored in nanoseconds, so we divide by 1,000,000 to get
+// milliseconds
+long get_duration_ms(const TimingInfo& info) {
+  return info.duration.count() / 1000000;
```
dalg24: Use `std::chrono::duration_cast`

profiling/energy-profiler/common/timer_system.cpp
```diff
@@ -0,0 +1,271 @@
...
+}
+
+// Helper function to format table columns with proper padding
+std::string format_table_cell(const std::string& content, size_t width) {
+  return content + std::string(width - std::min(content.size(), width), ' ');
+}
```
dalg24: You might want to lookup `std::setw`

profiling/energy-profiler/common/timer_system.hpp
```diff
@@ -0,0 +1,112 @@
...
+                       const char* src_name, const void* src_ptr,
+                       uint64_t size);
+  void end_deep_copy();
+
+  void push_profile_region(const char* region_name);
+  void pop_profile_region();
+
+  // Stack-based timing for robust region/kernel tracking
+  void start_region(const std::string& name, RegionType type, uint64_t id = 0);
+  void end_region();
```
dalg24: Fine but noting that if you can't correlate start and end of a region via some identifier then it won't be threadsafe.

tests/energy-profiler/test_csv_export.cpp
```diff
@@ -0,0 +1,72 @@
...
+  timer.end_region();
+
+  // Test the CSV export functions
+  const auto& kernels    = timer.get_kernel_timings();
+  const auto& regions    = timer.get_region_timings();
+  const auto& deepcopies = timer.get_deep_copy_timings();
+
+  // Expected: 2 kernels (ParallelFor and ParallelReduce), 1 region, 1 deep copy
+  ASSERT_EQ(kernels.size(), 2)
+      << "Expected 2 kernels, found " << kernels.size();
```
dalg24: You don't need to add this extra information, if it fails GTest will say kernels.sizeIO was not 2 and print the actual value.

tests/energy-profiler/test_csv_export.cpp
```diff
@@ -0,0 +1,72 @@
...
+      export_timings_csv(kernels, "test_kernels.csv", DataCategory::Kernels));
+  EXPECT_NO_THROW(
+      export_timings_csv(regions, "test_regions.csv", DataCategory::Regions));
+  EXPECT_NO_THROW(export_timings_csv(deepcopies, "test_deepcopies.csv",
+                                     DataCategory::DeepCopies));
+
+  // Test print functions (these should not throw)
+  EXPECT_NO_THROW(print_timings_summary(kernels, DataCategory::Kernels));
+  EXPECT_NO_THROW(print_timings_summary(regions, DataCategory::Regions));
+  EXPECT_NO_THROW(print_timings_summary(deepcopies, DataCategory::DeepCopies));
```
dalg24: That's not super useful testing that it does not throw.

tests/energy-profiler/test_timer_system.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+  const auto& timings = timer.get_kernel_timings();
+  ASSERT_EQ(timings.size(), 1) << "Expected 1 timing";
+
+  const auto& timing = timings[0];
+  uint64_t duration_ms =
+      std::chrono::duration_cast<std::chrono::milliseconds>(timing.duration)
+          .count();
+
+  EXPECT_TRUE(is_within_range(duration_ms, 2, 2))
```
dalg24: Do we really mean to have a tolerance that is that high?

profiling/energy-profiler/common/error_handling.hpp
```diff
@@ -0,0 +1,46 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos Energy Profiler Error Handling
```
dalg24: Do not modify the copyright headers

ethan-puyaubreau: solved

profiling/energy-profiler/common/timer_system.cpp
ethan-puyaubreau: indeed, replaced with `std::setw`, I forgot about its existence when refactoring the code.

profiling/energy-profiler/kokkos-tools/CMakeLists.txt
ethan-puyaubreau: that would definitely be something I would change to a envvar instead.

profiling/energy-profiler/kokkos-tools/kp_energy_kernel_timer.cpp
JBludau: But thinking about this some more, I think we should make the verbosity a runtime choice.

profiling/energy-profiler/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+# Energy Profiler Infrastructure
+# This provides the foundational infrastructure for energy profiling tools:
+# - Timer system for kernel and region tracking
+# - Common utilities (filename generation, error handling)
+# - Tool interface definitions
+# - Basic kernel timer tool
+
+add_subdirectory(kokkos-tools)
```
dalg24: This red circle means "no newline at end of file".
Please fix it by adding a newline character.
Same comment everywhere.

profiling/energy-profiler/common/filename_prefix.cpp
```diff
@@ -0,0 +1,31 @@
...
+//
+//@HEADER
+
+#include "filename_prefix.hpp"
+#include <unistd.h>
+
+namespace KokkosTools {
+namespace EnergyProfiler {
+
+std::string generate_prefix() {
```
dalg24, line 24:
```suggestion
std::string KokkosTools::EnergyProfiler::generate_prefix() {
```

profiling/energy-profiler/common/filename_prefix.hpp
```diff
@@ -0,0 +1,27 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#pragma once
+
+#include <string>
+
+namespace KokkosTools {
+namespace EnergyProfiler {
```
dalg24, line 24:
```suggestion
namespace KokkosTools::EnergyProfiler {
```

profiling/energy-profiler/kokkos-tools/kp_energy_kernel_timer.cpp
```diff
@@ -0,0 +1,236 @@
...
+// --- Core Timer Infrastructure ---
+EnergyProfiler::KernelTimerTool timer;
+
+// --- Library Initialization/Finalization ---
+
+void kokkosp_init_library(const int loadSeq, const uint64_t interfaceVer,
+                          const uint32_t devInfoCount,
+                          Kokkos_Profiling_KokkosPDeviceInfo* deviceInfo) {
+  // Enable verbose mode if environment variable is set
+  if (std::getenv("KOKKOS_TOOLS_ENERGY_VERBOSE")) {
```
dalg24: As currently implemented, if I do `KOKKOS_TOOLS_ENERGY_VERBOSE=0` or `=FALSE` it will set verbose mode.
Anyway I am not sure whether there is a convention in this repo at the moment.

tests/energy-profiler/test_csv_export.cpp
```diff
@@ -0,0 +1,69 @@
...
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#include <chrono>
+#include <thread>
+#include <string>
+#include "gtest/gtest.h"
+#include "../../profiling/energy-profiler/common/timer_system.hpp"
```
dalg24: Not sure I am ok with relative path from the test dir to the source.

profiling/energy-profiler/common/timer_system.cpp
```diff
@@ -0,0 +1,275 @@
...
+  (void)devInfoCount;
+  (void)deviceInfo;
+  std::cout << "Kokkos Energy Profiler: Initializing with load sequence "
+            << loadSeq << " and interface version " << interfaceVer << '\n';
+  std::cout << "Kokkos Energy Profiler: Library initialized" << '\n';
+}
+
+void KernelTimerTool::finalize_library() {
+  // Implementation is empty - resources cleaned up automatically by destructors
+}
```
dalg24: Why does the timer need an init() or a finalize()?
It adds unnecessary indirections and feels out of place.

profiling/energy-profiler/common/timer_system.cpp
```diff
@@ -0,0 +1,275 @@
...
+// Kokkos parallel operations
+void KernelTimerTool::begin_parallel_for(const char* name, uint32_t devID,
+                                         uint64_t& kID) {
+  (void)devID;
+  kID = next_region_id_++;
+  start_region(name, RegionType::ParallelFor, kID);
+}
+
+void KernelTimerTool::end_parallel_for(uint64_t kID) {
+  (void)kID;
```
dalg24:
```suggestion
void KernelTimerTool::end_parallel_for(uint64_t /*kID*/) {
```

tests/energy-profiler/CMakeLists.txt
ethan-puyaubreau: The test needs to verify internal implementation of the timer, I can make a separate shared library but that would require refactoring I think, or rethink the test so it uses the actual library and tests its behavior.

profiling/energy-profiler/timing/kp_energy_kernel_timer.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+    fprintf(file, "%s,%s,%ld,%ld,%ld\n", timing.name.c_str(), type.c_str(),
+            start_ms, end_ms, duration_ms);
+  }
+
+  fclose(file);
+  std::cout << "All timing data exported to " << filename << '\n';
+}
+
+void print_all_timings_summary(const std::deque<TimingInfo>& kernels,
```
JBludau: use std::ostream and an iterator. The function should only print one

profiling/energy-profiler/timing/kp_energy_kernel_timer.cpp
```diff
@@ -0,0 +1,376 @@
...
+                         });
+
+  if (it != g_active_regions.rend()) {
+    // Found the region with matching ID
+    auto region = *it;
+
+    // Remove the region from active list
+    g_active_regions.erase(std::next(it).base());
+
+    region.end_time = std::chrono::high_resolution_clock::now();
```
JBludau: move to top of function

profiling/energy-profiler/timing/kp_energy_kernel_timer.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+// Functions to be added to Kokkos library
+void kokkosp_init_library(const int loadSeq, const uint64_t interfaceVer,
+                          const uint32_t devInfoCount,
+                          Kokkos_Profiling_KokkosPDeviceInfo* deviceInfo) {
+  (void)devInfoCount;
+  (void)deviceInfo;
+
+  // Enable verbose mode if environment variable is set
+  if (std::getenv("KOKKOS_TOOLS_ENERGY_VERBOSE")) {
```
JBludau: would also be active for KOKKOS_TOOLS_ENERGY_VERBOSE=0

profiling/energy-profiler/timing_utils.hpp
```diff
@@ -0,0 +1,144 @@
...
+void end_region_by_type(RegionType type_to_end);
+void end_region_with_id(uint64_t expected_id);
+uint64_t generate_new_region_id();
+bool is_verbose_enabled();
+void set_verbose_enabled(bool enabled);
+void log_verbose(const char* format, ...);
+std::vector<TimingInfo> get_all_timings();
+
+/// @brief RAII class for automatic timing region management
+class ScopedTimingRegion {
```
JBludau: Hmm I am not sure why you added this at this state of the internship

profiling/energy-profiler/kp_energy_profiler.cpp
```diff
@@ -0,0 +1,313 @@
...
+#include <chrono>
+#include <cstdarg>
+#include <cstdint>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+#include <iostream>
+#include <mutex>
+#include <string>
+#include <vector>
```
JBludau, line 22: do we need all of them here?

profiling/energy-profiler/timing_utils.hpp
ethan-puyaubreau: I had already implemented something like this on the previous versions of the tool, so it was pretty easy to put here but I can remove it if needed.

profiling/energy-profiler/kp_energy_profiler.cpp
ethan-puyaubreau: Almost, the only one I have removed now is `<cstring>` since there's no more need for this one.
As for the rest in the file:
  - `<algorithm>`: `std::find_if`, `std::next`
  - `<chrono>`: `std::chrono::high_resolution_clock::now()`
  - `<cstdarg>`: `va_list`, `va_start`, `va_end`
  - `<cstdint>`: `uint64_t`
  - `<cstdio>`: `vprintf`
  - `<cstdlib>`: `std::getenv`
  - `<iostream>`: `std::cerr`, `std::cout`
  - `<mutex>`: `std::lock_guard`, `std::mutex`
  - `<string>`: `std::string`
  - `<vector>`: `std::vector`

profiling/energy-profiler/timing/kp_energy_kernel_timer.cpp
ethan-puyaubreau: Changed it so it only activates if `ON` or `1`

profiling/energy-profiler/timing_utils.hpp
ethan-puyaubreau: Removed it as for now it's not really useful, I'll keep it simple

profiling/energy-profiler/kp_energy_profiler.cpp
```diff
@@ -0,0 +1,316 @@
...
+  try {
+    auto end_time = std::chrono::high_resolution_clock::now();
+    auto& state   = EnergyProfilerState::get_instance();
+    std::lock_guard<std::mutex> lock(state.get_mutex());
+    auto& active_regions = state.get_active_regions();
+    if (active_regions.empty()) {
+      std::cerr << "Warning: Attempting to end region with ID " << expected_id
+                << " but no active regions found.\n";
+      return;
+    }
```
JBludau: the search would also raise this ... so maybe the precheck could be removed

profiling/energy-profiler/kp_energy_profiler.cpp
```diff
@@ -0,0 +1,316 @@
...
+                          Kokkos_Profiling_KokkosPDeviceInfo* deviceInfo) {
+  (void)devInfoCount;
+  (void)deviceInfo;
+  const char* verbose_env = std::getenv("KOKKOS_TOOLS_ENERGY_VERBOSE");
+  if (verbose_env &&
+      (std::string(verbose_env) == "1" || std::string(verbose_env) == "ON")) {
+    KokkosTools::EnergyProfiler::EnergyProfilerState::get_instance()
+        .set_verbose_enabled(true);
+  }
+  printf(
```
JBludau: please convert them to c++ and get rid of includes that are not required

profiling/energy-profiler/kp_energy_profiler.cpp
```diff
@@ -0,0 +1,307 @@
...
+    TimingInfo region;
+    region.name       = name;
+    region.type       = type;
+    region.start_time = std::chrono::high_resolution_clock::now();
+    region.id         = id;
+    auto& state       = EnergyProfilerState::get_instance();
+    std::lock_guard<std::mutex> lock(state.get_mutex());
+    state.get_active_regions().push_back(region);
+  } catch (const std::exception& e) {
+    std::cerr << "Error in start_region: " << e.what() << std::endl;
```
JBludau: hmm ... afaik that is not how exceptions are supposed to be used ... especially since you are not handling the exception but just printing and not rethrowing. So this would not lead to an abort.

if any of the functions in the try block throw, you are silencing that but also not restoring a valid state so that the program can continue. I would remove the try-catch

profiling/energy-profiler/kp_energy_profiler.cpp
```diff
@@ -0,0 +1,290 @@
...
+  auto& active_regions = state.get_active_regions();
+  if (active_regions.empty()) return;
+  auto it = std::find_if(active_regions.rbegin(), active_regions.rend(),
+                         [type_to_end](const TimingInfo& region) {
+                           return region.type == type_to_end;
+                         });
+  if (it != active_regions.rend()) {
+    auto region = *it;
+    active_regions.erase(std::next(it).base());
+    region.end_time = std::chrono::high_resolution_clock::now();
```
JBludau: You should take the time in all functions immediately (as first instruction). This way you don't pollute your timings with the time it takes to erase a region or the search. These might introduce errors of unknown order that is not even constant

profiling/energy-profiler/kp_energy_profiler.cpp
```diff
@@ -0,0 +1,290 @@
...
+    std::cout << message << std::endl;
+  }
+}
+
+// Start a region
+void start_region(const std::string& name, RegionType type, uint64_t id) {
+  TimingInfo region;
+  region.name       = name;
+  region.type       = type;
+  region.start_time = std::chrono::high_resolution_clock::now();
```
JBludau: when starting, you should take the time at the end and then use it to update the last region in the dqueue. This way you don't measure the construction or the lock

JBludau: As I am writing this: It of course does not hold for nested regions ... but that is too much for now

maartenarnst:
A recent reference with related work on rocm:
- https://escholarship.org/uc/item/3c11w9kg

vlkale:
> A recent reference with related work on rocm:
> 
> - https://escholarship.org/uc/item/3c11w9kg

Thanks!

PR #301, opened 2025-08-15
Energy profiling tools: NVML-based measurement tool

ethan-puyaubreau:
This PR adds NVML-based GPU energy monitoring to the Kokkos Energy Profiler:

Features:
- Integrated Energy Profiling: Combines timing measurements with GPU power monitoring in a single profiler library
- NVML Provider: `NVMLProvider` class for GPU power monitoring with device discovery and power usage APIs
- Background Power Sampling: `PowerSampler` class using daemon-based background monitoring for continuous power data collection
- Multi-GPU Support: Individual device queries for systems with multiple GPUs

Implementation:
- Main profiler library: `kp_energy_profiler` (integrates timing + power monitoring)
- NVML provider: `nvml_provider.cpp/hpp` for GPU power monitoring
- Power sampler: `power_sampler.cpp/hpp` for background sampling with daemon
- Timing utilities: Enhanced timing infrastructure with power data export

Dependencies:
- Requires CUDA 12.6+ with NVML support
- Depends on: #300 and #299

Usage:
The profiler automatically detects available GPUs and begins power monitoring during initialization. Power data is collected in the background and exported alongside timing data in CSV format.


profiling/energy-profiler/nvml/kp_nvml_power.cpp
```diff
@@ -0,0 +1,351 @@
...
+    devices_.resize(device_count);
+    device_names_.resize(device_count);
+    
+    for (unsigned int i = 0; i < device_count; ++i) {
+      nvml_result = nvmlDeviceGetHandleByIndex(i, &devices_[i]);
+      if (NVML_SUCCESS != nvml_result) {
+        std::cerr << "[KokkosPowerProfiler] WARNING: Failed to get handle for device " 
+                  << i << std::endl;
+        devices_[i] = nullptr;
+        device_names_[i] = "Failed Device";
```
JBludau: could be more descriptive

profiling/energy-profiler/nvml/kp_nvml_power.cpp
```diff
@@ -0,0 +1,351 @@
...
+    } else {
+      std::cerr << "[KokkosPowerProfiler] ERROR: Failed to get power usage for device " 
+                << device_index << ": " << nvmlErrorString(nvml_result) << std::endl;
+      return false;
+    }
+  }
+
+  size_t get_device_count() const { return devices_.size(); }
+  std::string get_device_name(size_t device_index) const {
+    if (device_index >= device_names_.size()) return "Unknown Device";
```
JBludau: could be more descriptive

profiling/energy-profiler/nvml/kp_nvml_power.cpp
```diff
@@ -0,0 +1,351 @@
...
+        device_names_[i] = "Failed Device";
+        continue;
+      }
+      
+      char device_name[NVML_DEVICE_NAME_BUFFER_SIZE];
+      nvml_result = nvmlDeviceGetName(devices_[i], device_name, NVML_DEVICE_NAME_BUFFER_SIZE);
+      if (NVML_SUCCESS == nvml_result) {
+        device_names_[i] = std::string(device_name);
+      } else {
+        device_names_[i] = "Unknown Device " + std::to_string(i);
```
JBludau: Maybe "NVML could not discover device i"

profiling/energy-profiler/nvml/kp_nvml_power.cpp
```diff
@@ -0,0 +1,351 @@
...
+      }
+      
+      // Test power usage reading
+      unsigned int test_power_mW = 0;
+      nvml_result = nvmlDeviceGetPowerUsage(devices_[i], &test_power_mW);
+      if (NVML_SUCCESS != nvml_result) {
+        std::cerr << "[KokkosPowerProfiler] WARNING: Device " << i 
+                  << ": Power usage reading failed: " << nvmlErrorString(nvml_result) << std::endl;
+      }
```
JBludau: That could be removed

profiling/energy-profiler/nvml/kp_nvml_power.cpp
```diff
@@ -0,0 +1,351 @@
...
+    }
+    return true;
+  }
+  
+  bool validate_device_index(size_t device_index) const {
```
JBludau: this is kind of redundant

profiling/energy-profiler/nvml/kp_nvml_power.cpp
```diff
@@ -0,0 +1,351 @@
...
+    if (device_index >= devices_.size()) {
+      std::cerr << "[KokkosPowerProfiler] ERROR: Device index " << device_index << " out of range" << std::endl;
+      return false;
+    }
+    return true;
+  }
+};
+
+// --- Power Profiler Core Logic ---
+std::string generate_prefix() {
```
JBludau: use the other prefix you already have

profiling/energy-profiler/nvml/kp_nvml_power.cpp
```diff
@@ -0,0 +1,351 @@
...
+void kokkosp_finalize_library() {
+  if (g_power_daemon) {
+    g_power_daemon->stop();
+  }
+
+  std::vector<PowerSample> samples_copy;
+  {
+    std::lock_guard<std::mutex> lock(g_samples_mutex);
+    samples_copy = g_power_samples;
+    g_power_samples.clear(); // Clear the global vector for the next run
```
JBludau: There is no second run, right?

CMakeLists.txt
```diff
@@ -103,6 +103,23 @@ else()
   set(KOKKOSTOOLS_HAS_VTUNE OFF)
 endif()
 
+# Check for NVML (required for energy profiler)
+set(MIN_CUDA_VERSION 12.6)
+find_package(CUDAToolkit ${MIN_CUDA_VERSION} QUIET)
+if (CUDAToolkit_FOUND)
```
romintomasetti, line 109: Have you seen https://docs.nvidia.com/nsight-systems/UserGuide/index.html#nvml-power-and-temperature-metrics-preview ?

It would be great to actually better document this tool and compare it to what `nsigh-systems` may provide. Does your tool make anything special to help correlate `Kokkos` regions with consumption, trigger anything special ? Or is it "just" launching a thread that samples the energy consumption ?

romintomasetti: Also, what about asynchronicity ? I mean:
```c++
{
    Kokkos::Profiling::ScopeRegion region("my region");

    Kokkos::parallel_for(Kokkos::RangePolicy(exec, 0, N), ...); // async
}
```

If the tool reports the consumption of the *my region* region, if you're not `Kokkos::fence`ing, what is the meaning of the measurements reported by `Kokkos Tools`, especially when the kernel is actually running *after* the scoped region ends ?

romintomasetti (changes requested):
I would not merge this PR without an example.

And I would be strongly against merging this one before we get proper `Cuda` testing, *e.g.* #271.

CMakeLists.txt
ethan-puyaubreau: Hello! I've seen this page one time but since I focused more on Variorum support at some point I haven't had the chance of reading it further. As of now, the tool is only launching a thread that samples the energy consumption.

ethan-puyaubreau: > Also, what about asynchronicity ? I mean:
> ```c++
> {
>     Kokkos::Profiling::ScopeRegion region("my region");
> 
>     Kokkos::parallel_for(Kokkos::RangePolicy(exec, 0, N), ...); // async
> }
> ```
> 
> If the tool reports the consumption of the *my region* region, if you're not `Kokkos::fence`ing, what is the meaning of the measurements reported by `Kokkos Tools`, especially when the kernel is actually running *after* the scoped region ends ?

For now, the system doesn't do anything special to correlate power consumption with a specific region or kernel. The timing system is meant to help visualize what the current situation is, so there's room for improvement (meaning adding more correlation using multiple metrics), especially since the daemon system would still allow for method/data propagation

JBludau: @romintomasetti we will check if Nsight does actually do something useful. It at least sounds like the exact same thing we are doing. 
About the fencing: Nvidia at the moment allows power measurements every 100ms. But it seems to return only the power average of the last 25ms of that window, see http://arxiv.org/abs/2312.02741. Thus the tool will at the current state of Nvidas tools only be useful for measuring entire regions and even then it will need repetitions with shifts and post processing in order to get anything that is relatable to an algorithm. Due to these problems the tool currently does not require fencing, this should be done by the user.

maartenarnst: Could be good to add the AMD tool to the mix:
- https://github.com/ROCm/rocprofiler-systems?tab=readme-ov-file#gpu-metrics

JBludau: on our ToDo list :-)

vlkale: @JBludau Do you still have plans to look at the AMD Profiling metrics, as mentioned above? Do you have updates you can share - particularly those that are pertinent to this PR - from your end? 

Thanks!

JBludau: This pr has been split into smaller ones to make it easier to review. Once we have these in, we can think about adding something for amd

PR #303, opened 2025-08-29
Update ci c++20

JBludau:
Proposes to use c++20 since we require it in Kokkos and the CI here uses the latest develop. Not sure if we want to keep it that way

### DIFF
--- a/.github/workflows/build-with-kokkos.yml
+++ b/.github/workflows/build-with-kokkos.yml
@@ -13,13 +13,13 @@ jobs:
           - image: ubuntu:22.04
             preset: OpenMP
             compiler: default
-          - image: nvidia/cuda:12.1.0-devel-ubuntu22.04
+          - image: nvidia/cuda:12.2.0-devel-ubuntu22.04
             preset: Cuda
             compiler: default
           - image: nvidia/cuda:12.2.0-devel-ubuntu22.04
             preset: Cuda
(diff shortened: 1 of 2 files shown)


dalg24 (approved):
Fine by me if it passes

masterleinad (approved):

.github/workflows/build-with-kokkos.yml
```diff
@@ -13,13 +13,13 @@ jobs:
...
+          - image: nvidia/cuda:12.2.0-devel-ubuntu22.04
             preset: Cuda
             compiler: default
           - image: nvidia/cuda:12.2.0-devel-ubuntu22.04
             preset: Cuda
             compiler: {cpp: g++-12, c: gcc-12}
-          - image: rocm/dev-ubuntu-22.04:5.4
+          - image: rocm/dev-ubuntu-22.04:6.3
             preset: ROCm
             compiler: default
```
maartenarnst, line 24: This case now appears to duplicate the case that follows.

Also, it is probably important to doublecheck with @Rombur, because in the past, I think we had to cover the range of versions that is used on Frontier.

dalg24: Uggg.  Thanks for pointing that out.

JBludau: Oh, good catch. Yeah, we should also think about what we want to keep testing while 4 and 5 are both supported

PR #308, opened 2025-10-13
Fix suspicious value of `BUILD_SHARED_LIBS`

tpadioleau:
The PR addresses a suspicious assignment of `BUILD_SHARED_LIBS` to a list containing `Build shared libraries;ON`. I guess it was meant to assign the value `ON`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,7 +29,7 @@ message(STATUS Configuring Kokkos-Tools)
 message(STATUS)
 
 # Common settings
-set(BUILD_SHARED_LIBS "Build shared libraries" ON)
+set(BUILD_SHARED_LIBS ON)
 if(WIN32)
   set(BUILD_SHARED_LIBS OFF) # We need to add __declspec(dllexport/dllimport) for Windows DLLs
 endif()


CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ message(STATUS Configuring Kokkos-Tools)
 message(STATUS)
 
 # Common settings
-set(BUILD_SHARED_LIBS "Build shared libraries" ON)
+set(BUILD_SHARED_LIBS ON)
```
dalg24, line 32: It looks more like a mistake where the author meant to use `option(BUILD_SHARED_LIBS "Build shared libraries" ON)` which seems to be a rather common pattern.
It is mentioned in the cmake tutorials.  See https://cmake.org/cmake/help/latest/guide/tutorial/Selecting%20Static%20or%20Shared%20Libraries.html

masterleinad: I'm pretty sure the author intended to always build shared libraries.

masterleinad (approved):

CMakeLists.txt
dalg24: Assume I configure with `-D BUILD_SHARED_LIBS=OFF` on Linux.
Why is it OK to ignore the request and overwrite the variable (assuming that it actually works when `BUILD_SHARED_LIBS` becomes a cache variable)?

masterleinad: We want to always build shared libraries so we can `dlopen` them. Using a local variable will shadow the CACHE variable (and thus have the intended behavior). I see that esentially as a shortcut to setting the library type explicitly for every library (which we could do instead).

dalg24: I think it is generally bad form to silently ignore a setting requested by the user but I recognize that the PR in its current form is an improvement over the status quo.
I will wait a few more hours to give the opportunity for others to chime in and I will merge if no one requests changes.

tpadioleau: @dalg24 I admit it is a possiblity, to be honest I directly thought the intent was to enforce shared libraries. Note that in the case of Windows the cache value is ignored because a normal variable is set.

JBludau (approved):
Definitely an improvement. But it certainly is confusing. If we always want to build a shared lib, we should specify that on `add_library` explicitly

tpadioleau:
> If we always want to build a shared lib, we should specify that on `add_library` explicitly

I am confused with the Windows case, I don't know if we can do that.

masterleinad:
> I am confused with the Windows case, I don't know if we can do that.

Why not? I mean Windows support for Kokkos Tools needs work anyway but I don't fundamental issues.

CMakeLists.txt
crtrott: Note that default absolutely, always not necessarily. It is valid to statically link a tool libs into your app (in particular when building as a monolithic library) and then use the API interfaces to choose which tool functions to load. For the monolithic case each individual tool is putting its functions into a namespace do make that work.

tpadioleau: I was missing this use case, I can rework the PR to make it a cache variable with a default value that depends on the platform. That way we get back control from the command line (and Spack) and have a reasonable default for plugins.

dalg24: Let's do it as a follow up PR

vlkale: > Note that default absolutely, always not necessarily. It is valid to statically link a tool libs into your app (in particular when building as a monolithic library) and then use the API interfaces to choose which tool functions to load. For the monolithic case each individual tool is putting its functions into a namespace do make that work. 

Yes, dynamic libraries is the most common use case / scenario; the option to build a static library (.a for Linux, lib for Windows) should not be eliminated. 

The change proposed is fine for now (and thanks @tpadioleau !), but users really should not be forced into building tool as a dynamic library.

PR #309, opened 2025-10-13
Introduce options to disable Variorum and SystemTap

tpadioleau:
The PR adds two CMake options to disable Variorum (`KokkosTools_ENABLE_VARIORUM`) and SystemTap (`KokkosTools_ENABLE_SYSTEMTAP`). This work is related to the issue https://github.com/spack/spack-packages/issues/377 that needs an explicit way to disable the search of Variorum and dtrace.


cmake/configure_variorum.cmake
```diff
@@ -21,5 +21,5 @@ find_package(Variorum QUIET)
 if(Variorum_FOUND)
   set(KOKKOSTOOLS_HAS_VARIORUM TRUE)
 else()
-  message(WARNING "Variorum not found: ${MSG_NOTFOUND}")
+  message(FATAL_ERROR "Variorum not found: ${MSG_NOTFOUND}")
```
tpadioleau, line 24: The message is made `FATAL_ERROR` because the user can now explicitly disable Variorum if it cannot be found.

I think once a new release of variorum is made, shipping https://github.com/LLNL/variorum/commit/c3b620ab00f72ce2fb0b75aee46ce1319b86a13d, we could replace this file by `find_package(Variorum REQUIRED)`.

CMakeLists.txt
```diff
@@ -157,13 +161,12 @@ if(KokkosTools_ENABLE_PAPI)
...
-  find_program(KOKKOSTOOLS_DTRACE_EXECUTABLE dtrace)
-  if(KOKKOSTOOLS_DTRACE_EXECUTABLE)
+if(KokkosTools_ENABLE_SYSTEMTAP)
+  if(NOT WIN32 AND NOT APPLE)
+    find_program(KOKKOSTOOLS_DTRACE_EXECUTABLE dtrace REQUIRED)
     add_subdirectory(profiling/systemtap-connector)
     set(KOKKOSTOOLS_HAS_SYSTEMTAP ON)
   else()
-    message(STATUS "Skipping systemtap-connector (dtrace executable wasn't found)")
     set(KOKKOSTOOLS_HAS_SYSTEMTAP OFF)
```
dalg24: If the user explicitly ask to enable, we should not silently turn it OFF, we should at the very least raise a warning, preferably an error.

cmake/configure_variorum.cmake
dalg24: Why don't we instead do `find_package(Variorum REQUIRED)` L19 and drop that if...else statement.

dalg24 (commented):
Adding options is fine because it is consistent with what we currently do with other dependency.

Longer-term, I would like us to consider simply using the CMake native support for controlling finding packages, `CMAKE_DISABLE_FIND_PACKAGE_<PackageName>` and `CMAKE_REQUIRE_FIND_PACKAGE_<PackageName>`, instead of rolling our own.
(Not blocking your PR though as it is a legitimate improvement over the status quo)

cmake/configure_variorum.cmake
tpadioleau: My intent was to keep the message to help the user to set the expected variables. For example the user could expect that setting `CMAKE_PREFIX_PATH` works.

CMakeLists.txt
tpadioleau: I agree, what do you think of failing in this case and telling the user this option is not supported on those platforms ?

tpadioleau:
> Longer-term, I would like us to consider simply using the CMake native support for controlling finding packages, `CMAKE_DISABLE_FIND_PACKAGE_<PackageName>` and `CMAKE_REQUIRE_FIND_PACKAGE_<PackageName>`, instead of rolling our own. (Not blocking your PR though as it is a legitimate improvement over the status quo)

A difficulty I see with CMake options is that they only relate to dependencies but not to features of the project. Some features have multiple dependencies, some dependencies belong to different features (like MPI). If we remove all our options and rely only on the predefined variables, I think it might end up be more difficult for the user to find the correct options to set.

CMakeLists.txt
crtrott: Yeah

tpadioleau: done in 6b11729207f2edb125fb9b0d340acc0c68689703

cmake/configure_variorum.cmake
tpadioleau: Any strong preference ?

CMakeLists.txt
```diff
@@ -157,14 +161,13 @@ if(KokkosTools_ENABLE_PAPI)
   add_subdirectory(profiling/papi-connector)
 endif()
 
-if(NOT WIN32 AND NOT APPLE)
-  find_program(KOKKOSTOOLS_DTRACE_EXECUTABLE dtrace)
-  if(KOKKOSTOOLS_DTRACE_EXECUTABLE)
+if(KokkosTools_ENABLE_SYSTEMTAP)
+  if(NOT WIN32 AND NOT APPLE)
```
dalg24, line 165: I would slightly prefer avoiding the negations
```cmake
if(WIN32 OR APPLE)
  message(FATAL_ERROR ...
else()
  find_package(...

cmake/configure_variorum.cmake
dalg24: > My intent was to keep the message to help the user to set the expected variables. For example the user could expect that setting `CMAKE_PREFIX_PATH` works.

Why would it not work?

> Any strong preference ?

No.  That's fine we can rework the find modules later as necessary.

dalg24 (approved):

CMakeLists.txt
tpadioleau: Good point

cmake/configure_variorum.cmake
tpadioleau: > Why would it not work?
> 
> >

Variorum used to install the CMake files in a non standard directory, CMake could not find them with `CMAKE_PREFIX_PATH` nor `Variorum_ROOT`. I suspect this is why we have all the logic about paths in the `configure_variorum.cmake`. The commit https://github.com/LLNL/variorum/commit/c3b620ab00f72ce2fb0b75aee46ce1319b86a13d should fix that but this will be available only in the next release of Variorum.

PR #317, opened 2026-03-09
Add version number and export it into `C++` and `CMake`

JBludau:
Not sure what the version number should be right now. But I guess it makes sense to keep it up to date with the ecosystem.


vlkale:
This should probably be version 0.x or 1.x. There is not a version number for Kokkos Tools right now or when David was developing it.

cwpearson (approved):
Good with it once we settle on a version number

dalg24:
> This should probably be version 0.x or 1.x. There is not a version number for Kokkos Tools right now or when David was developing it.

No we are going to do lockstep with Kokkos Core and Kokkos Kernels.

vlkale:
> > This should probably be version 0.x or 1.x. There is not a version number for Kokkos Tools right now or when David was developing it.
> 
> No we are going to do lockstep with Kokkos Core and Kokkos Kernels.

This sounds even better to me.

dalg24 (approved):

PR #327, opened 2026-03-11
Also test with latest release

masterleinad:
Setting up for whatever range of versions we want to support in the end.

### DIFF
--- a/.github/workflows/build-with-kokkos.yml
+++ b/.github/workflows/build-with-kokkos.yml
@@ -9,10 +9,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        include:
+        kokkos-version-tag: [5.0.2, develop]
+        build:
           - image: ubuntu:22.04
             preset: OpenMP
             compiler: default
+
           - image: nvidia/cuda:12.2.0-devel-ubuntu22.04
             preset: Cuda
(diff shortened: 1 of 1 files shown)


.github/workflows/build-with-kokkos.yml
```diff
@@ -9,10 +9,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        include:
+        kokkos-version-tag: [5.0.2, develop]
```
dalg24, line 12: I don't think we should test against develop in the CI but that belongs to nightlies or weeklies instead

vlkale (commented):
Do we also want to test for backward compatibility with earlier versions of Kokkos, say, Kokkos 4.5? 

I agree at minimum a latest major release version of Kokkos (Kokkos 5.x right now) ought to be CI tested against Kokkos Tools weekly and even nightly.

.github/workflows/build-with-kokkos.yml
masterleinad: That's pre-existing at least.

JBludau: I agree, I would take our latest 5 and latest 4 since we kind of promised long term support

PR #336, opened 2026-07-08
SpaceTimeStack: Expose deep_copy

masterleinad:
Right now, we don't track any calls to `deep_copy`. This pull request fixes that.


dalg24 (approved):
It looks like an oversight in the big #159 overhaul

PR #341, opened 2026-07-22
CMAKE_CXX_EXTENSIONS OFF for builds with nvcc_wrapper

janciesko:
Two changes;
- Force strict C++20 for CUDA builds with nvcc_wrapper otherwise we get the warning [1]
- Only export defined Kokkos variables from acquire_kokkos_config if they were defined

[1] warning
```bash
nvcc_wrapper has been given GNU extension standard flag -std=gnu++20 - reverting flag to -std=c++20
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,6 +63,18 @@ option(KokkosTools_ENABLE_TESTS    "Build tests"                    OFF)
 
 # Fetch Kokkos options:
 acquire_kokkos_config()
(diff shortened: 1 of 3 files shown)


masterleinad:
Why do we want to force this unconditionally and not just when building with `nvcc` or for `Cuda`?

janciesko:
@masterleinad  Fixed

crtrott (approved):

PR #343, opened 2026-07-27
Makes Windows build fail earlier with more descriptive error message 

janciesko:
Reopens https://github.com/kokkos/kokkos-tools/pull/298


dalg24:
Please provide a description that explains why we went from partial support to outright preventing to build on Windows.

PR #348, opened 2026-09-15
Enforce the CMake C++ standard requirement

masterleinad:
Compiling with gcc-14 didn't work because we just used `-std=c++1z` and couldn't find the `filesystem` header.


CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ endif()
 if(CMAKE_CXX_STANDARD LESS 20)
   message(FATAL_ERROR "KokkosTools requires C++20")
 endif()
-
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
JBludau, line 22: Why not add `CXX_STANDARD_REQUIRED` to the Kokkos target?

masterleinad: Which target do you mean?

JBludau: kokkoscore

masterleinad: We are only linking with `Kokkos::kokkos` in the tests; the tools libraries don't dependent on `kokkoscore`.

JBludau: Ok, I misunderstood the intend then

JBludau (approved):

